PR #3, opened 2016-06-21
Feature/run ps script

LaviniaC:


UiPath.Script.Activities.Design/DesignerMetadata.cs
```diff
@@ -12,13 +14,27 @@ public class DesignerMetadata : IRegisterMetadata
         public void Register()
         {
             var builder = new AttributeTableBuilder();
+
+            var categoryAttribute = new CategoryAttribute("App Scripting");
             var autoHotKeyType = typeof(ExecuteAutoHotKey);
```
vladjerca: I think we should change the definition to "RunAutoHotKeyScript", for naming consistency

PR #19, opened 2018-05-07
Added the EnumerateObjects activity.

LeifLazarUiPath:


florinakke:
We should update the package minor version since a new activity was added

FTP/UiPath.FTP/FtpSession.cs
```diff
@@ -153,6 +153,36 @@ private void _ftpClient_ValidateCertificate(FtpClient control, FtpSslValidationE
             return listing;
         }
 
+        private IEnumerable<FtpObjectInfo> GetRemoteListing(string remotePath, bool recursive)
+        {
+            if (string.IsNullOrWhiteSpace(remotePath))
+            {
+                throw new ArgumentNullException(nameof(remotePath));
```
cosmyo: This is not localized. I think it should be `Resources.RemotePath`

LeifLazarUiPath: remotePath is a method parameter, the ArgumentNullException composes the message itself.

PR #20, opened 2018-05-24
Update GetNumberOfColumnsFromRange in order to support columns after AA

DanielaDumitrascu:
Issue Description: Read Range with range columns after AA return "Read range : Input array is longer than the number of columns in this table". 
The changes I made seem to work, but should be also tested more and it needs to be improved  (like the name of the columns in the range could be lower case also? if so, the regular expression be updated)

### DIFF
--- a/Activities/Google Spreadsheet/GoogleSpreadsheet.Activities.UnitTest/ReadRangeTest.cs
+++ b/Activities/Google Spreadsheet/GoogleSpreadsheet.Activities.UnitTest/ReadRangeTest.cs
@@ -37,5 +37,24 @@ namespace GoogleSpreadSheetUnitTests
             Assert.AreEqual(4, numberOfRows);
         }
 
+        [TestMethod]
+        public void GetNumberOfColumnsFromRangeAfterColumnZTest1()
+        {
+            var range = "A3:AD7";
+
+            var numberOfColumns = ReadRange.GetNumberOfColumnsFromRange(range);
+
+            Assert.AreEqual(30, numberOfColumns);
+        }
+
+        [TestMethod]
+        public void GetNumberOfColumnsFromRangeAfterColumnZTest2()
+        {
+            var range = "Z3:AF7";
+
+            var numberOfColumns = ReadRange.GetNumberOfColumnsFromRange(range);
+
+            Assert.AreEqual(7, numberOfColumns);
+        }
     }
 }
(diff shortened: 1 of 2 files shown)


Google Spreadsheet/GoogleSpreadsheet.Activities/ReadRange.cs
```diff
@@ -94,14 +95,58 @@ public static int GetNumberOfColumnsFromRange(string range)
...
             }
 
-            return secondLetter - firstLetter + 1;
+            return secondColNumber - firstColNumber + 1;
+
+            //return secondLetter - firstLetter + 1;
         }
 
+
+        public static int letterToColumn(String letter)
```
bogdan-popescu-uipath: Please make sure that current unit tests are still passing and add new ones for the added functionality.

Google Spreadsheet/GoogleSpreadsheet.Activities/ReadRange.cs
```diff
@@ -94,14 +95,58 @@ public static int GetNumberOfColumnsFromRange(string range)
...
 
-            if (secondLetter < firstLetter)
+            /* if (secondLetter < firstLetter)
+             {
+                 throw new Exception("Invalid range specified.");
+             }*/
+
+
+            // Handle also columns after AA
+            Regex regexRange = new Regex(@"([A-Z]+)\d+");
```
adrian-zanescu: We could initialize the regex as a static member and compile it to reuse it between calls

Google Spreadsheet/GoogleSpreadsheet.Activities/ReadRange.cs
adrian-zanescu: maybe LetterToColumn for C# consistency

Google Spreadsheet/GoogleSpreadsheet.Activities/ReadRange.cs
```diff
@@ -94,14 +95,58 @@ public static int GetNumberOfColumnsFromRange(string range)
...
+            string firstColName, secondColName;
+            int firstColNumber = 0, secondColNumber = 0;
+            if (match.Success)
+            {
+                firstColName = match.Groups[1].Value;
+                firstColNumber = letterToColumn(firstColName);
+            }
+            else
+            {
+                throw new Exception("Invalid range specified.");
```
adrian-zanescu: We should try to throw more specific exception types like ArgumentException or InvalidOperation

PR #21, opened 2018-06-04
Google Spreadsheet Write range: change InputValueOption from RAW to U…

DanielaDumitrascu:
Issue Description: In the spreadsheets the numeric values or dates are written as strings with a leading apostrophe. The input should be parsed exactly as if it were entered into the Google Sheets UI  so "Mar 1 2016" should be a date. https://developers.google.com/sheets/api/guides/values.


LaviniaC:
possible breaking change if someone actually used the activity and expected to write string instead of actual values. @adrian-zanescu what do you think? are these activities out in the wild?

adrian-zanescu:
not sure if and how much they are used but the change should be fine.  it will not make any existing workflow to fail and the differences in output are minor and beneficial

PR #53, opened 2019-09-01
added PrivateKeyAuthenticationMethod to SftpSession

IlyaKochetov:
This is to fix an error with SFTP session when the authentication has to be done via RSA key, not the password. Current code just ignored the existence of CertificatePath argument, always trying to connect using password and causing an error "`String reference not set to an instance of a String, parameter name s`" from SSH.NET lib if password was not supplied.
Also added a description to CertificatePath argument suggesting that the valid format for the key is PEM PCKS#1


theloginerror:
This pull request has been mentioned on **UiPath Community Forum**. There might be relevant details there:

https://forum.uipath.com/t/error-when-trying-to-connect-to-a-sftp-server-with-a-private-key-and-without-password/193655/9

theloginerror:
This pull request has been mentioned on **UiForum Sandbox**. There might be relevant details there:

https://sandbox-forum.uipath.com/t/added-privatekeyauthenticationmethod-to-sftpsession-pr-53/614/1

VarchasvTyagi:
Guys, can we compile and push these changes to the NuGet feed?
Multiple users are trying to upgrade to this version and there is a concern about using the same within the Production environments.

PR #69, opened 2020-10-15
Python.Activities: Fix threading model for load/execute script

florinakke:
Some modules ( like xlwings) perform COM initialization ( https://docs.microsoft.com/en-us/windows/win32/api/combaseapi/nf-combaseapi-coinitializeex ); this fails since the threading model is already set on MTA. Ensure that Load/Exec script are executed on STA threads.

### DIFF
--- a/Activities/Python/UiPath.Python/Impl/Engine.cs
+++ b/Activities/Python/UiPath.Python/Impl/Engine.cs
@@ -124,7 +124,7 @@ namespace UiPath.Python.Impl
         public Task<PythonObject> LoadScript(string code, CancellationToken ct)
         {
             ct.ThrowIfCancellationRequested();
-            return Task.Run(() =>
+            return RunSTA(() =>
             {
                 using (_py.GIL())
                 {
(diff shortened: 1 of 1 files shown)


cosmyo (approved):
Tested with the project provided internally.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -124,7 +124,7 @@ public Task Release()
         public Task<PythonObject> LoadScript(string code, CancellationToken ct)
         {
             ct.ThrowIfCancellationRequested();
-            return Task.Run(() =>
+            return RunSTA((cancellationToken) =>
```
LeifLazarUiPath, line 127: No need to pass ct to the func, it can use cancellationToken from the caller.

LeifLazarUiPath (changes requested):

Activities/Python/UiPath.Python/Impl/Engine.cs
florinakke: fixed; also combined the commits to a single one.

LeifLazarUiPath (approved):

grigoremircea:
Deployed to Connect Feed (https://gallery.uipath.com/api/v2) as version "1.0.7594.23824"

grigoremircea:
Correction deployed as version v1.1.7594.35963 on Connect feed.

theloginerror:
This pull request has been mentioned on **UiPath Community Forum**. There might be relevant details there:

https://forum.uipath.com/t/load-python-script-error-when-running-python-script/200446/14

PR #80, opened 2020-12-18
OSAC-62 - database driver errors

alcxs:
+ Description
DB activities catch the driver errors and re-throw them wrapped in a more RPA-dev-friendly:
"[Database driver error]: <original error message>"
+ How to test
Create a new WF
Add a database connection
Configure SQL Server / Oracle details
Shutdown the database service (to cause the error)
+ Jira
[OSAC-62](https://uipath.atlassian.net/browse/OSAC-62)
+ Label:
24S

### DIFF
--- a/Activities/Database/UiPath.Database.Activities/DatabaseConnect.cs
+++ b/Activities/Database/UiPath.Database.Activities/DatabaseConnect.cs
@@ -1,6 +1,7 @@
 ﻿using System;
 using System.Activities;
 using System.ComponentModel;
+using System.Data.Common;
 using System.Runtime.CompilerServices;
 using System.Windows.Markup;
 using UiPath.Database.Activities.Properties;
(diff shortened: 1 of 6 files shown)


Activities/Database/UiPath.Database.Activities/DatabaseConnect.cs
```diff
@@ -40,19 +41,33 @@ internal DatabaseConnect(IDBConnectionFactory factory)
...
+                var provName = ProviderName.Get(context);
+                Func<DatabaseConnection> action = () => _connectionFactory.Create(connString, provName);
+                context.UserState = action;
 
-            return action.BeginInvoke(callback, state);
+                return action.BeginInvoke(callback, state);
+            }
+            catch (DbException ex)
+            {
+                throw new Exception("[Database driver error]: " + ex.Message + " " + ex?.InnerException?.Message, ex);
```
adrian-zanescu, line 55: possible breaking change for a workflow that catches DbException. We should not change exception types. And if we throw we should be a bit more specific than Exception

adrian-zanescu (changes requested):

PR #84, opened 2020-12-28
Added parameter mapping for out parameters for "Execute query" activity

petcua1:
Until now only for Execute Non Query out parameters were mapped. In SQL Server a stored procedure can return a result set as well as out parameters


adrian-zanescu (approved):

adrian-zanescu (changes requested):
we have some unit tests failing

PR #95, opened 2021-02-03
OSAC 63 - Java tests

alcxs:
Description
Integration tests for Java activities
How to test
Run Azure DevOps pipeline
Jira
[OSAC-63](https://uipath.atlassian.net/browse/OSAC-63)
Label:
24S


catalin-marin (commented):
can we have just one references.db ? I see one in python and one for java

alcxs:
> can we have just one references.db ? I see one in python and one for java

The two files mentioned are from different test suites (independent UiPath projects). Recommendation would be to keep the projects separate, as they do not have common activities and are needed to test different activity packages. This is why is better to have two references.db files.

catalin-marin (approved):

PR #125, opened 2021-03-23
OSAC 128 - Prepare Credentials activities for localization

alcxs:
Description
Prepare Credentials activities for localization
How to test
Change Studio language and check translations for Credentials activities
Jira
[OSAC-128](https://uipath.atlassian.net/browse/OSAC-128)
Label:
24S


Activities/Credentials/UiPath.Credentials.Activities/AddCredential.cs
```diff
@@ -1,28 +1,43 @@
...
 using System.ComponentModel;
+using System;
+using System.Activities.Validation;
+using System.Diagnostics;
+using System.IO;
+using UiPath.Credentials.Activities.Properties;
 
 namespace UiPath.Credentials.Activities
 {
-    [Category("System.Credentials")]
```
adrian-zanescu, line 7: this is the place in tree where the activities are placed by studio after load. Not sure it's safe to remove it. We possibly need a LocalizedCategory attribute

adrian-zanescu (changes requested):

Activities/Credentials/UiPath.Credentials.Activities/AddCredential.cs
alcxs: As dicussed, changes added to DesignerMetadata in latest commit.

adrian-zanescu (approved):

PR #142, opened 2021-06-15
OSAC 164 - [Cryptography] Insecure unauthenticated encryption in Encrypt Text

alcxs:
Description
Encrypt Text - is using an antiquated mode of AES encryption, one that does not authenticate the message
How to test
load Credential package and use AES GCM algorithm
Jira
[OSAC-164](https://uipath.atlassian.net/browse/OSAC-164)
Label:
24S


Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -45,74 +54,117 @@ public static byte[] HashDataWithKey(KeyedHashAlgorithms keyedHashAlgorithm, byt
...
             {
-                byte[] encrypted;
-                byte[] salt = new byte[PBKDF2_SaltSizeBytes];
-                int maxKeySize = GetLegalKeySizes(algorithm).Max();
-
-                _rng.GetBytes(salt);
-                using (Rfc2898DeriveBytes pbkdf2 = new Rfc2898DeriveBytes(key, salt, PBKDF2_Iterations))
+                return EncryptAesGcmNet461(inputBytes, key, out result);
+            }
+            else
```
ovidiuiliescu: Please refactor this to be more readable, as the condition itself is the same in both cases. E.g.

```
if (symmetricAlgorithm == SymmetricAlgorithms.AESGCM)
{
#if NET5_0
...
#else
...
#endif
}
```

Same elsewhere in the code (e.g in the decrypt methods).

Please also use don't explicitly check for NET5_0 and NET461, as the targets may change later and we won't go back through the code to replace them, resulting in potential issues.

Instead, use `NETCOREAPP` and `NETFRAMEWORK` as they are more flexible, see the table at the end of https://docs.microsoft.com/en-us/dotnet/standard/frameworks#how-to-specify-a-target-framework .

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -45,74 +54,117 @@ public static byte[] HashDataWithKey(KeyedHashAlgorithms keyedHashAlgorithm, byt
...
-                Buffer.BlockCopy(algorithm.IV, 0, result, salt.Length, algorithm.IV.Length);
-                Buffer.BlockCopy(encrypted, 0, result, salt.Length + algorithm.IV.Length, encrypted.Length);
+                return result;
+#if NET461
             }
-
-            return result;
+#endif
+#if NET5_0
+            }
```
ovidiuiliescu, line 96: This kind of preprocessing is really bad practice and is not needed. Having a single closing bracket as the start of a preprocessing region is a strong code smell.

In this particular case, you don't need to have the code like this:

```
if (...) return A
else if (...) return B
else return C
```

You can just have

```
if (...) return A

if (...) return B

return C
```

It makes the code much more readable when preprocessing directives are involved.

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -260,5 +315,145 @@ private static int[] GetLegalKeySizes(SymmetricAlgorithm algorithm)
...
+            using (Rfc2898DeriveBytes pbkdf2 = new Rfc2898DeriveBytes(key, salt, PBKDF2_Iterations))
+            {
+                var Key = pbkdf2.GetBytes(32); //256 bit key
+                decrypted = DecryptWithBouncyCastle(encryptedData, iv, tag, Key);
+            }
+
+            return decrypted;
+        }
+
+        private static byte[] EncryptWithBouncyCastle(byte[] plaintext, byte[] key, byte[] nonce, out byte[] tag)
```
ovidiuiliescu: Let's rename this to `EncryptAesGcmWithBouncyCastle` or similar, so that future us will know what algorithm the code uses.

Same for decrypt.

Activities/Database/Database.build.props
```diff
@@ -2,6 +2,6 @@
   <PropertyGroup>
     <Company>UiPath</Company>
     <Copyright>© UiPath</Copyright>
-    <VersionPrefix>1.5.0</VersionPrefix>
+    <VersionPrefix>1.6.0</VersionPrefix>
```
ovidiuiliescu: Why are these changed here ?
Same for other package versions.

ovidiuiliescu (changes requested):

ovidiuiliescu (commented):
We also need some unit tests around this new functionality.

Ideally cross-check .NET 5 with BouncyCastle and vice-versa.

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -260,5 +315,145 @@ private static int[] GetLegalKeySizes(SymmetricAlgorithm algorithm)
...
+
+            result = new byte[salt.Length + algorithmIV.Length + encrypted.Length + tag.Length];
+            Buffer.BlockCopy(salt, 0, result, 0, salt.Length);
+            Buffer.BlockCopy(algorithmIV, 0, result, salt.Length, algorithmIV.Length);
+            Buffer.BlockCopy(encrypted, 0, result, salt.Length + algorithmIV.Length, encrypted.Length);
+            Buffer.BlockCopy(tag, 0, result, salt.Length + algorithmIV.Length + encrypted.Length, tag.Length);
+
+            return result;
+        }
+        private static byte[] DecryptAesGcmNet5(byte[] inputBytes, byte[] key, out byte[] decrypted)
```
ovidiuiliescu: Nitpick: please insert an empty line before this method.

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -222,7 +274,10 @@ private static SymmetricAlgorithm GetSymmetricAlgorithm(SymmetricAlgorithms symm
             switch (symmetricAlgorithm)
             {
                 case SymmetricAlgorithms.AES:
-                    return new AesCryptoServiceProvider(); // TODO: Use AesCng after upgrading to .NET Framework 4.6.2
+                    return new AesCryptoServiceProvider(); // kept for backwords compat
+                case SymmetricAlgorithms.AESGCM:
```
ovidiuiliescu, line 260: Nitpick: Empty line before this

Also before the default below.

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -260,5 +315,145 @@ private static int[] GetLegalKeySizes(SymmetricAlgorithm algorithm)
 
             return keySizes.ToArray();
         }
+
+#if NET5_0
+        private static byte[] EncryptAesGcmNet5(byte[] inputBytes, byte[] key, out byte[] result)
```
ovidiuiliescu: There's a lot of overlap between this method and the 461 one.

Same for the decryption counterparts.

Can we extract the common parts into dedicated methods and reuse them?

Or, since it seems that only the part inside the `using`s is different, maybe create some helper methods around those and have a single set of Encrypt/Decrypt methods, which call the appropriate helper methods inside them via `#ifdef` ?

Something like

```
private static byte[] EncryptAesGcm(...)
{
// Init

using()
{
#if NET5_0
EncryptAesGcmWith50(...);
#endif

#if NET461
EncryptAesGcmWith461(...);
#endif
}

// Return result
}
```

This also has the advantage of removing a lot of the preprocessing directives elsewhere, since we have a common entry point method for both NET 5.0 and NET 461, and we only need preprocessing inside that method.

Either way, we need to cut down on the code duplication.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: changed to match recommendation

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: changed, as requested

Activities/Database/Database.build.props
alcxs: renamed

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: added

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: added

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: modified

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: renamed

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -46,68 +55,96 @@ public static byte[] EncryptData(SymmetricAlgorithms symmetricAlgorithm, byte[]
...
-                byte[] salt = new byte[PBKDF2_SaltSizeBytes];
-                int maxKeySize = GetLegalKeySizes(algorithm).Max();
+#if NETCOREAPP
 
-                _rng.GetBytes(salt);
-                using (Rfc2898DeriveBytes pbkdf2 = new Rfc2898DeriveBytes(key, salt, PBKDF2_Iterations))
-                {
-                    algorithm.Key = pbkdf2.GetBytes(maxKeySize);
-                }
+                return EncryptAesGcm(inputBytes, key, out result);
```
ovidiuiliescu: Why is the same method called in both branches?

If we call the same code in both cases, let's remove the ifdefs.

Same for the DecryptData method.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
alcxs: sure, removed.

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -260,5 +298,124 @@ private static int[] GetLegalKeySizes(SymmetricAlgorithm algorithm)
 
             return keySizes.ToArray();
         }
+
+        private static byte[] EncryptAesGcm(byte[] inputBytes, byte[] key, out byte[] result)
```
ovidiuiliescu: Any good reason why we have result as an out argument and also as a return value?
Can we have it just as a return value?

Same for decrypt method.

alcxs: removed as out parameter

PR #143, opened 2021-06-17
OSAC 164 - fixed descriptions and default value

alcxs:
Description
Encrypt Text - is using an antiquated mode of AES encryption, one that does not authenticate the message
How to test
load Credential package and use AES GCM algorithm
Jira
[OSAC-164](https://uipath.atlassian.net/browse/OSAC-164)
Label:
24S

### DIFF
--- a/Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
@@ -58,7 +58,7 @@ namespace UiPath.Cryptography.Activities
 
         public DecryptFile()
         {
-            Algorithm = SymmetricAlgorithms.AES;
+            Algorithm = SymmetricAlgorithms.AESGCM;
             KeyEncoding = new VisualBasicValue<Encoding>(typeof(Encoding).FullName + "." + nameof(Encoding.UTF8)); // Kinda ugly.
         }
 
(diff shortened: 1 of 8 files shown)


Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -19,7 +19,7 @@ public class DecryptFile : CodeActivity
         [LocalizedCategory(nameof(Resources.Input))]
         [LocalizedDisplayName(nameof(Resources.AlgorithmDisplayName))]
         [LocalizedDescription(nameof(Resources.DecryptAlgorithmDescription))]
-        public SymmetricAlgorithms Algorithm { get; set; }
+        public SymmetricAlgorithms Algorithm { get; set; } = SymmetricAlgorithms.AESGCM;
```
ovidiuiliescu: It should be enough to set this in the constructor. Setting it in both places is a bit redundant and might lead to harder to understand code.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
alcxs: removed, per request

ovidiuiliescu (approved):

PR #181, opened 2021-09-27
[OSC-241] Bug fixes on FTP on Linux migration

petcua1:


Activities/Shared/UiPath.Shared.Activities/ContinuableAsyncNativeActivity.cs
```diff
@@ -16,7 +17,7 @@ protected override void Execute(NativeActivityContext context)
             }
             catch (Exception e)
             {
-                if (ContinueOnError.Get(context))
+                if (continueOnError)
```
ovidiuiliescu: A better solution would be to define `ContinueOnError` as abstract in this class.

Then all derived classes would be forced to implement it properly.

See how this is done in the main Activities repository.

ovidiuiliescu (changes requested):

Activities/Shared/UiPath.Shared.Activities/ContinuableAsyncNativeActivity.cs
petcua1: fixed

ovidiuiliescu (approved):

ovidiuiliescu (approved):

PR #214, opened 2022-01-13
OSAC-270, OSAC-328 - Execute Query throws error at runtime that Value…

radunegulescu2:
… Cannot be null even activity is valid in Studio Web

### DIFF
--- a/Activities/Database/UiPath.Database.Activities/ConnectionHelper.cs
+++ b/Activities/Database/UiPath.Database.Activities/ConnectionHelper.cs
@@ -14,7 +14,7 @@ namespace UiPath.Database.Activities
         {
             if (existingConnection == null && connString == null && connSecureString == null && provName == null)
             {
-                throw new ArgumentNullException(Resources.ValidationError_ConnectionMustNotBeNull);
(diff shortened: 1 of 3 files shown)


Activities/Database/UiPath.Database.Activities/ConnectionHelper.cs
```diff
@@ -30,7 +30,7 @@ public static void ConnectionValidation(DatabaseConnection existingConnection, S
                 {
                     throw new ArgumentException(Resources.ValidationError_ConnectionStringMustBeSet);
                 }
-                if (connString == null && existingConnection == null)
+                if (connString == null && connSecureString == null)
```
petcua1, line 33: What happens if existing connection is not null? It will not work with existing connection

petcua1 (changes requested):

Activities/Database/UiPath.Database.Activities/ConnectionHelper.cs
radunegulescu2: That if is inside another if: if (existingConnection == null), so the existingConnection can't be not null there.

alcxs (approved):

ovidiuiliescu (approved):

PR #238, opened 2022-03-07
OSAC-272 - Implement IRemoteResource for Cryptography package activities

alcxs:
https://uipath.atlassian.net/browse/OSAC-272


Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -69,6 +70,22 @@ public partial class DecryptFile : CodeActivity
...
+        [LocalizedDisplayName(nameof(Resources.Activity_DecryptFile_Property_InputFile_Name))]
+        [LocalizedDescription(
+            nameof(Resources.Activity_DecryptFile_Property_InputFile_Description))]
+        public InArgument<IResource> InputFile { get; set; }
+
+        [Browsable(false)]
+        [DefaultValue(null)]
+        [LocalizedCategory(nameof(Resources.Output))]
+        [LocalizedDisplayName(nameof(Resources.Activity_DecryptFile_Property_DecryptedFile_Name))]
+        [LocalizedDescription(nameof(Resources
```
ovidiuiliescu: Why 2 lines here, resource keys should be on line line even if they are longer.

Also see comments left in the view model.

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -126,20 +144,39 @@ protected override void Execute(CodeActivityContext context)
...
-                byte[] encrypted = File.ReadAllBytes(inputFilePath);
-
-                byte[] decrypted = null;
-                try
+                if (inputFile != null && !inputFile.IsFolder)
                 {
-                    decrypted = CryptographyHelper.DecryptData(Algorithm, encrypted, CryptographyHelper.KeyEncoding(keyEncoding, key, keySecureString));
+                    // Get local file
+                    var localFile = inputFile.ToLocalResource();
+                    Task.Run(async () => { await localFile.ResolveAsync(); });
```
ovidiuiliescu: This will not wait for the resolve to actually happen. Please either:

a) derive this from an async actvitiy, see how it's done in the Activities repo.
b) do Task.Run(..).GetAwaiter().GetResult() to make sure we actually wait for the resolve to complete. You can use `ResolveAndGetUnderlyingFile` for this or something similar, to make sure we don't needlessly duplicate code.

Same in other places this happens.

Activities/Cryptography/UiPath.Cryptography.Activities/Models/CryptographyLocalItem.cs
```diff
@@ -0,0 +1,119 @@
...
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using UiPath.Cryptography.Activities.Extensions;
+using UiPath.Cryptography.Activities.Properties;
+using UiPath.Platform.ResourceHandling;
+
+namespace UiPath.Cryptography.Activities.Models
+{
+    public class CryptographyRemoteItem : IRemoteResource
```
ovidiuiliescu: This should never actually be used, since crypto activities have no concept of "remote" files.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptFileViewModel.cs
```diff
@@ -75,19 +66,30 @@ public DecryptFileViewModel(IDesignServices services) : base(services)
...
 
+        /// <summary>
+        /// The file that you want to encrypt.
+        /// </summary>
+        public DesignInArgument<IResource> InputFile { get; set; } = new DesignInArgument<IResource>();
+
+        /// <summary>
+        /// The file that you want to encrypt.
+        /// </summary>
+        public DesignOutArgument<CryptographyRemoteItem> DecryptedFile { get; set; } = new DesignOutArgument<CryptographyRemoteItem>();
```
ovidiuiliescu: This should be `ILocalResource` since the file is now locally available AND there's nothing crypto-specific about it (it's just a local resource like any other from a consumer's perspective, no extra bells and whistles about it).

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptFileViewModel.cs
```diff
@@ -75,19 +66,30 @@ public DecryptFileViewModel(IDesignServices services) : base(services)
         /// </summary>
         public DesignInArgument<bool> ContinueOnError { get; set; } = new DesignInArgument<bool>();
 
+        /// <summary>
+        /// The file that you want to encrypt.
+        /// </summary>
+        public DesignInArgument<IResource> InputFile { get; set; } = new DesignInArgument<IResource>();
+
+        /// <summary>
+        /// The file that you want to encrypt.
```
ovidiuiliescu: Comment is wrong.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptFileViewModel.cs
```diff
@@ -75,18 +66,33 @@ public EncryptFileViewModel(IDesignServices services) : base(services)
...
 
+        /// <summary>
+        /// The file that you want to encrypt.
+        /// </summary>
+        public DesignInArgument<IResource> InputFile { get; set; } = new DesignInArgument<IResource>();
+
+        /// <summary>
+        /// The file that you want to encrypt.
+        /// </summary>
+        public DesignOutArgument<CryptographyRemoteItem> EncryptedFile { get; set; } = new DesignOutArgument<CryptographyRemoteItem>();
```
ovidiuiliescu: See comments left for the decript view model.

Activities/Cryptography/UiPath.Cryptography.Activities/EncryptFile.cs
```diff
@@ -36,52 +46,72 @@ public partial class EncryptFile : CodeActivity
...
+            nameof(Resources.Activity_EncryptFile_Property_InputFile_Description))]
+        public InArgument<IResource> InputFile { get; set; }
+
+        [Browsable(false)]
+        [DefaultValue(null)]
+        [LocalizedCategory(nameof(Resources.Output))]
+        [LocalizedDisplayName(nameof(Resources.Activity_EncryptFile_Property_EncryptedFile_Name))]
+        [LocalizedDescription(nameof(Resources
+            .Activity_EncryptFile_Property_EncryptedFile_Description))]
+        public OutArgument<CryptographyRemoteItem> EncryptedFile { get; set; }
```
ovidiuiliescu: See comments left in the view model.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
alcxs: resolved

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
alcxs: resolved

Activities/Cryptography/UiPath.Cryptography.Activities/Models/CryptographyLocalItem.cs
alcxs: removed

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptFileViewModel.cs
alcxs: resolved

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptFileViewModel.cs
alcxs: resolved

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptFileViewModel.cs
alcxs: resolved

Activities/Cryptography/UiPath.Cryptography.Activities/EncryptFile.cs
alcxs: resolved

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -126,20 +142,35 @@ protected override void Execute(CodeActivityContext context)
                     throw new ArgumentException(Resources.FileAlreadyExistsException, Resources.OutputFilePathDisplayName);
                 }
 
-                byte[] encrypted = File.ReadAllBytes(inputFilePath);
-
-                byte[] decrypted = null;
-                try
+                if (inputFile != null && !inputFile.IsFolder)
```
ovidiuiliescu, line 147: These ifs should be consolidated.

Ideally you would get a local file path via ifs, but after you get a local file path then the code should be the same.

Likewise, the output should be handled for both cases, e.g. if I have both the output path and the output IResource properties set, I should write to both.

Same for other places in this PR where this applies.

Activities/Cryptography/UiPath.Cryptography.Activities/Models/CryptographyLocalItem.cs
```diff
@@ -0,0 +1,79 @@
...
+    /// Used to convert cryptography files as IResources.
+    /// </summary>
+    public class CryptographyLocalItem : ILocalResource
+    {
+        /// <summary>
+        /// Create the ILocalResource implementation
+        /// </summary>
+        /// <param name="byteArray"></param>
+        /// <param name="fullName"></param>
+        internal CryptographyLocalItem(byte[] byteArray, string fullName)
```
ovidiuiliescu, line 20: This isn't lazy. Ideally the decryption when using IResources should only happen when you call .Resolve() on this item, not when the item is created.

This being said, if it's extremely complicated to achieve this, the current behavior is OK for now (or at least until someone complains).

Activities/Cryptography/UiPath.Cryptography.Activities/Models/CryptographyLocalItem.cs
```diff
@@ -0,0 +1,79 @@
...
+        public DateTime? CreationDate => _fileInfo?.CreationTime;
+
+        /// <inheritdoc />
+        public string FullName { get; }
+
+        /// <inheritdoc />
+        public string IconUri => string.Empty;
+
+        /// <inheritdoc />
+        public string ID => Guid.NewGuid().ToString();
```
ovidiuiliescu: This will generate a new ID every time the property is accessed. Wrong on so many levels.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptFileViewModel.cs
```diff
@@ -119,6 +117,8 @@ protected override void InitializeModel()
                 .AddMenuProperty(Key, KeyInputMode.Key)
                 .AddMenuProperty(KeySecureString, KeyInputMode.SecureKey)
                 .BuildAndInsertMenuActions();
+
+            DecryptedFile.OrderIndex = propertyOrderIndex++;
```
ovidiuiliescu, line 122: Is this really all we need to set here ? We should at least set explicity the IsPrincipal and IsVisible values.

Same for other similar occurrences ?

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/Models/CryptographyLocalItem.cs
ovidiuiliescu: Hm, I think you need to decrypt immediately in the activity anyway if you want to output to a file as well (not just to a resource).

I think we need to clarify with the PM what the desired behavior is if you have set both the IResource output and the file output. Based on that discussion, this may be OK going forward and we can introduce a new "lazy" constructor when the need arises.

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -69,6 +70,20 @@ public partial class DecryptFile : CodeActivity
         [LocalizedDescription(nameof(Resources.Activity_DecryptFile_Property_ContinueOnError_Description))]
         public InArgument<bool> ContinueOnError { get; set; }
 
+        [Browsable(false)]
+        [DefaultValue(null)]
+        [LocalizedCategory(nameof(Resources.Input))]
+        [LocalizedDisplayName(nameof(Resources.Activity_DecryptFile_Property_InputFile_Name))]
+        [LocalizedDescription(nameof(Resources.Activity_DecryptFile_Property_InputFile_Description))]
+        public InArgument<IResource> InputFile { get; set; }
```
ovidiuiliescu, line 78: We need to either document what takes priority between this and the file path as string, or throw an exception if both are set to prevent confusion.

Please take this up with product. Also this impacts other activities I think.

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
```diff
@@ -92,34 +99,44 @@ protected override string Execute(CodeActivityContext context)
...
-                }
 
+                byte[] hashed;
+
+                if (inputFile != null && !inputFile.IsFolder)
+                {
+                    // Get local file
+                    var localFile = inputFile.ToLocalResource();
+                    //Run Sync
+                    Task.Run(async () => await localFile.ResolveAsync()).GetAwaiter().GetResult();
```
ovidiuiliescu, line 136: These ifs still need consolidating. No need to call CryptographyHelper.HashDataWithKey both in the if and in the else branch, refactor so that it's called only once after the if..else block. See how it's already done in this PR for other activities.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
alcxs: As discussed with Product, an exception is raised when both are filled in. Text message of exception is validated with Product as well.

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
alcxs: fixed

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -90,43 +105,54 @@ protected override void Execute(CodeActivityContext context)
...
-                if (string.IsNullOrWhiteSpace(outputFilePath))
-                {
+
+                if (string.IsNullOrWhiteSpace(outputFilePath) && inputFile == null)
                     throw new ArgumentNullException(Resources.OutputFilePathDisplayName);
-                }
+
+                //either input file path or input file as resource should be used
+                if (!string.IsNullOrWhiteSpace(inputFilePath) && inputFile != null)
+                    throw new ArgumentNullException(string.Format(Resources.Exception_UseOnlyFilePathOrInputResource,
```
ovidiuiliescu: `ArgumentNullException` is misleading, as it's only supposed to be used when an argument is null.
Let's throw plain `ArgumentException` instead.

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
```diff
@@ -92,34 +99,46 @@ protected override string Execute(CodeActivityContext context)
...
-                if (!File.Exists(filePath))
-                {
+
+                if (!File.Exists(filePath) && inputFile == null)
                     throw new ArgumentException(Resources.FileDoesNotExistsException, Resources.FilePathDisplayName);
-                }
 
+                byte[] hashed;
+
+                if (inputFile != null && !inputFile.IsFolder)
```
ovidiuiliescu, line 131: Shouldn't we throw an exception or something if the input file is a folder?

Same for other activities as well.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
alcxs: fixed

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
alcxs: done

ovidiuiliescu (approved):

PR #287, opened 2022-09-01
OSAC-393 - Rename "with Key" from the the Hash Text with Key and Hash…

radunegulescu2:
… File with Key activities


alcxs (changes requested):
add them to previous PR, it's too little to create a PR. add all renamings to same PR

PR #298, opened 2022-09-26
OSAC-387 - Consolidate Hash Text/File and Hash Text/File With Key act…

radunegulescu2:
…ivities

### DIFF
--- a/Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
@@ -6,6 +6,7 @@ using System.ComponentModel;
 using System.Diagnostics;
 using System.IO;
 using System.Security;
+using System.Security.Cryptography;
 using System.Text;
 using System.Threading.Tasks;
 using UiPath.Cryptography.Activities.Properties;
@@ -45,7 +46,7 @@ namespace UiPath.Cryptography.Activities
         [LocalizedDescription(nameof(Resources.Activity_KeyedHashFile_Property_KeySecureString_Description))]
         public InArgument<SecureString> KeySecureString { get; set; }
 
-        [RequiredArgument]
+        [Browsable(false)]
         [LocalizedCategory(nameof(Resources.Input))]
         [LocalizedDisplayName(nameof(Resources.Activity_KeyedHashFile_Property_Encoding_Name))]
         [LocalizedDescription(nameof(Resources.Activity_KeyedHashFile_Property_Encoding_Description))]
@@ -84,6 +85,20 @@ namespace UiPath.Cryptography.Activities
                 var error = new ValidationError(Resources.FipsComplianceWarning, true, nameof(Algorithm));
                 metadata.AddValidationError(error);
             }
+            if (Algorithm.ToString().StartsWith(nameof(HMAC)))
+            {
+                if (Key == null && KeyInputModeSwitch == KeyInputMode.Key)
+                {
+                    var error = new ValidationError(Resources.KeyNullError, false, nameof(Key));
+                    metadata.AddValidationError(error);
+                }
+                if (KeySecureString == null && KeyInputModeSwitch == KeyInputMode.SecureKey)
+                {
+                    // the validation error is added for having the both fields validated  
+                    var error = new ValidationError(Resources.KeySecureStringNullError, false, nameof(KeySecureString));
+                    metadata.AddValidationError(error);
+                }
+            }
 #if NET461
             if (Algorithm == KeyedHashAlgorithms.MACTripleDES)
             {
@@ -108,16 +123,22 @@ namespace UiPath.Cryptography.Activities
                 if (string.IsNullOrWhiteSpace(filePath) && inputFile == null)
                     throw new ArgumentNullException(Resources.FilePathDisplayName);
 
-                if (string.IsNullOrWhiteSpace(key) && keySecureString == null)
-                    throw new ArgumentNullException(Resources.KeyAndSecureStringNull);
+                if (Algorithm.ToString().StartsWith(nameof(HMAC)))
+                {
+                    if (string.IsNullOrWhiteSpace(key) && KeyInputModeSwitch == KeyInputMode.Key)
+                    {
+                        throw new ArgumentNullException(Resources.Activity_KeyedHashText_Property_Key_Name);
+                    }
+                    if ((keySecureString == null || keySecureString?.Length == 0) && KeyInputModeSwitch == KeyInputMode.SecureKey)
+                    {
+                        throw new ArgumentNullException(Resources.Activity_KeyedHashText_Property_KeySecureString_Name);
+                    }
+                }
 
                 //either input file path or input file as resource should be used
                 if (!string.IsNullOrWhiteSpace(filePath) && inputFile != null)
                     throw new ArgumentException(string.Format(Resources.Exception_UseOnlyFilePathOrInputResource,
                         Resources.Activity_KeyedHashFile_Property_InputFile_Name, Resources.Activity_KeyedHashFile_Property_FilePath_Name));
-
-                if (key != null && keySecureString != null)
-                    throw new ArgumentNullException(Resources.KeyAndSecureStringNotNull);
 
                 if (keyEncoding == null)
                     throw new ArgumentNullException(Resources.Encoding);
--- a/Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashText.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashText.cs
@@ -5,9 +5,11 @@ using System.Activities.Validation;
 using System.ComponentModel;
 using System.Diagnostics;
 using System.Security;
+using System.Security.Cryptography;
 using System.Text;
 using UiPath.Cryptography.Activities.Properties;
 using UiPath.Cryptography.Enums;
+using UiPath.Studio.Activities.Api.Activities;
 
 namespace UiPath.Cryptography.Activities
 {
@@ -43,7 +45,7 @@ namespace UiPath.Cryptography.Activities
         [LocalizedDescription(nameof(Resources.Activity_KeyedHashText_Property_KeySecureString_Description))]
         public InArgument<SecureString> KeySecureString { get; set; }
 
-        [RequiredArgument]
+        [Browsable(false)]
         [LocalizedCategory(nameof(Resources.Input))]
         [LocalizedDisplayName(nameof(Resources.Activity_KeyedHashText_Property_Encoding_Name))]
         [LocalizedDescription(nameof(Resources.Activity_KeyedHashText_Property_Encoding_Description))]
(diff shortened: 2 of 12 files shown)


Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +88,70 @@ protected override void InitializeModel()
             KeySecureString.IsVisible = false;
             KeySecureString.OrderIndex = propertyOrderIndex++;
 
-            KeyInputModeSwitch.IsVisible = false;
+            var useKeyMenuAction = new MenuAction
+            {
+                DisplayName = "Use Key",
```
alcxs: add this as resources for localization - or check if there are exising resources already

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +88,70 @@ protected override void InitializeModel()
...
+                Handler = UseKeyHandler
+            };
+
+            var useKeySecureStringMenuAction = new MenuAction
+            {
+                DisplayName = "Use Key Secure String",
```
alcxs: same here - localize

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashTextViewModel.cs
```diff
@@ -86,54 +86,69 @@ protected override void InitializeModel()
...
-            KeyInputModeSwitch.IsVisible = false;
+            var useKeyMenuAction = new MenuAction
+            {
+                DisplayName = "Use Key",
+                Handler = UseKeyHandler
+            };
+
+            var useKeySecureStringMenuAction = new MenuAction
+            {
+                DisplayName = "Use Key Secure String",
```
alcxs: localize

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashTextViewModel.cs
```diff
@@ -86,54 +86,69 @@ protected override void InitializeModel()
             KeySecureString.IsVisible = false;
             KeySecureString.OrderIndex = propertyOrderIndex++;
 
-            KeyInputModeSwitch.IsVisible = false;
+            var useKeyMenuAction = new MenuAction
+            {
+                DisplayName = "Use Key",
```
alcxs: localize

Activities/Cryptography/UiPath.Cryptography.Activities/UiPath.Cryptography.Activities.csproj
```diff
@@ -63,6 +64,7 @@
 		<Folder Include="Resources\" />
 	</ItemGroup>
 	<ItemGroup>
-		<Compile Condition="'$(TargetFramework)' != 'net5.0'" Remove="NetCore/**" />
+		<Compile Condition="'$(TargetFramework)' != 'net6.0'" Remove="NetCore/**" />
 	</ItemGroup>
+	<ProjectExtensions><VisualStudio><UserProperties resources_4activitiesmetadata_1json__JsonSchema="https://download.stackhawk.com/hawk/jsonconfig/hawkconfig.json" /></VisualStudio></ProjectExtensions>
```
alcxs: what is this? why is is needed?

alcxs (changes requested):
see individual comments

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +89,70 @@ protected override void InitializeModel()
...
+            Key.AddMenuAction(useKeySecureStringMenuAction);
+            KeySecureString.AddMenuAction(useKeyMenuAction);
 
             Result.IsPrincipal = false;
             Result.OrderIndex = propertyOrderIndex++;
 
             ContinueOnError.IsPrincipal = false;
             ContinueOnError.OrderIndex = propertyOrderIndex++;
-            ContinueOnError.Widget = new DefaultWidget { Type = ViewModelWidgetType.NullableBoolean };
+            ContinueOnError.Widget = new DefaultWidget { Type = "NullableBoolean", Metadata = new Dictionary<string, string>() };
```
ovidiuiliescu: Why do we need to hardcode this string ?

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +89,70 @@ protected override void InitializeModel()
...
-        /// <summary>
-        /// Key input Mode has changed. Set controls visibility based on selection
-        /// </summary>
-        private void KeyInputModeChanged_Action()
+        /// <inheritdoc/>
+        protected override void InitializeRules()
         {
-            switch (KeyInputModeSwitch.Value)
+            base.InitializeRules();
+            Rule("VisibleKey", () =>
```
ovidiuiliescu: Can we use `nameof(PropertyName)` instead of hardcoding strings here ?

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
```diff
@@ -84,6 +78,16 @@ protected override void CacheMetadata(CodeActivityMetadata metadata)
...
                 metadata.AddValidationError(error);
             }
+            if (Algorithm.ToString().StartsWith("HMAC") && Key == null && KeySecureString == null)
+            {
+                var error1 = new ValidationError(Resources.KeyAndSecureStringNull, false, nameof(Key));
+                metadata.AddValidationError(error1);
+
+                // error message is blank for not duplicating the error message
+                // the validation error is added for having the both fields validated  
+                var error2 = new ValidationError("", false, nameof(KeySecureString));
```
ovidiuiliescu: `string.Empty` instead of `""`

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashText.cs
```diff
@@ -75,6 +69,16 @@ protected override void CacheMetadata(CodeActivityMetadata metadata)
                 var error = new ValidationError(Resources.FipsComplianceWarning, true, nameof(Algorithm));
                 metadata.AddValidationError(error);
             }
+            if (Algorithm.ToString().StartsWith("HMAC") && Key == null && KeySecureString == null)
+            {
+                var error1 = new ValidationError(Resources.KeyAndSecureStringNull, false, nameof(Key));
```
ovidiuiliescu: Not sure what this block does and why we need to add a blank validation error.

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashFile.cs
```diff
@@ -84,6 +78,16 @@ protected override void CacheMetadata(CodeActivityMetadata metadata)
                 var error = new ValidationError(Resources.FipsComplianceWarning, true, nameof(Algorithm));
                 metadata.AddValidationError(error);
             }
+            if (Algorithm.ToString().StartsWith("HMAC") && Key == null && KeySecureString == null)
```
ovidiuiliescu: `HMAC` should be a const somewhere, it's not a good idea to have it as a magic string in code, especially since it's reused.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashText.cs
radunegulescu2: This block adds the validation on the Key field, if the field is empty and the algorithm starts with "HMAC". The blank validation is used for having the validation shown on the KeySecureString field also, but without duplicating the message.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
radunegulescu2: It's the name of the rule defined there, I don't think there are variables which can be put there for naming that rule.

ovidiuiliescu: How this should be handled is as follows (or at least how it's handled in Office and other activities).

* Each property should have its own rule (e.g. for key, for key secure string, etc)
* Each rule should have its own handler method
* In the handler, you can call the same private method or methods to update other properties. The idea is that any change on the value of property A that affects properties B, C, and D should be resolved by having the handler of property A call "UpdateStateOfB", "UpdateStateOfC" etc

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +89,70 @@ protected override void InitializeModel()
...
+                Handler = UseKeyHandler
+            };
+
+            var useKeySecureStringMenuAction = new MenuAction
+            {
+                DisplayName = Resources.Activity_KeyedHashFile_Property_KeySecureString_SwitchName,
+                Handler = UseKeySecureStringHandler
+            };
+
+            _useKeySecureString = KeySecureString.Value != null;
```
ovidiuiliescu: Using this private member to keep track of state based on the value of a given property initially, but THEN updating it via menu handlers feels wrong and is error prone.

Why not just keep the dedicated input mode property? It would make the code a whole lot cleaner and easier to track/understand.

It's also a pattern we have on other activities, where the mode is explicitly captured in the activity state/XAML.

You can also use this information to only show the validation errors for the current mode, e,g. no need to generate/show a validation error for secure string if we're not in secure string mode - and this works in both the CacheMetadata and ViewModel layers.

ovidiuiliescu (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities/KeyedHashText.cs
```diff
@@ -60,6 +54,8 @@ public partial class KeyedHashText : CodeActivity<string>
         [LocalizedDescription(nameof(Resources.Activity_KeyedHashText_Property_ContinueOnError_Description))]
         public InArgument<bool> ContinueOnError { get; set; }
 
+        private const string HMAC = "HMAC";
```
ovidiuiliescu: Just a hint, you can use `nameof(HMAC)` here instead of the string "HMAC".

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +89,66 @@ protected override void InitializeModel()
...
-        /// <summary>
-        /// Key input Mode has changed. Set controls visibility based on selection
-        /// </summary>
-        private void KeyInputModeChanged_Action()
+        /// <inheritdoc/>
+        protected override void InitializeRules()
         {
-            switch (KeyInputModeSwitch.Value)
+            base.InitializeRules();
+            Rule(nameof(Key), () =>
```
ovidiuiliescu: Let's have these as private methods instead in inline lambdas.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/KeyedHashFileViewModel.cs
```diff
@@ -89,54 +89,66 @@ protected override void InitializeModel()
             KeySecureString.IsVisible = false;
             KeySecureString.OrderIndex = propertyOrderIndex++;
 
-            KeyInputModeSwitch.IsVisible = false;
+            var useKeyMenuAction = new MenuAction
```
ovidiuiliescu: The actual user selection (key vs secure key) is not persisted in the activity. This can lead to the following scenario:

* I add the activity fresh, it's in "Key" mode.
* I select "Secure Key" mode.
* I close the project or refresh the page
* Upon reopen, the activity will still be in "Key" mode.

This needs to be discussed with the PM, if we want this behavior. For other activities, the user selected state is clearly captured in the activity (would be the equivalent of `KeyInputModeSwitch` that was removed).

ovidiuiliescu (changes requested):

ovidiuiliescu (approved):

alcxs (approved):

PR #383, opened 2023-08-24
Update WPF designers and descriptions STUD-65656

IoanDanielSavu:
## Description
DesignerMetadata is currently duplicating properties and activities descriptions. This leads to properties that have different placeholders from descriptions. At the same time, some activities designers are missing properties names.

## Jira
https://uipath.atlassian.net/browse/STUD-65656

image


andrei-balint (commented):
applies to all 3 PRs: there are many minor changes, also some refactorings in the class hierarchy.
What is missing is test infrastructure that checks for each activity and its view model the expected property values. For all targets, each with its specifics (designer metadata vs activitiesmetadata.json). Then the PR changes could be applied. 
I know this might be out of scope, but this is my general observation.

alexandru-petre (approved):

alexandru-petre:
> applies to all 3 PRs: there are many minor changes, also some refactorings in the class hierarchy. What is missing is test infrastructure that checks for each activity and its view model the expected property values. For all targets, each with its specifics (designer metadata vs activitiesmetadata.json). Then the PR changes could be applied. I know this might be out of scope, but this is my general observation.

Created separate task: https://uipath.atlassian.net/browse/STUD-65773

PR #424, opened 2024-04-10
Database: Windows OLEDB driver is not working (24.6) [STUD-64430]

viogroza:
Remove SNI loading workaround (since dotnet runtime bug was fixed https://github.com/dotnet/runtime/issues/16048) 
Remove native SNI dlls from package
Fix buffer overruns when configuring Connect to Database activity using Connection Wizard


viogroza:
the sonar issue reported about "_isWindows" is a false positive since the flag is used for NETCOREAPP to add additional options

alexandru-petre (approved):

PR #434, opened 2024-07-18
FTP: Upgrade package references (24.11) [STUD-71020]

viogroza:
Upgrade SSH.NET to 2024.0.0
Upgrade FluentFTP to 50.1.0
Small refactoring due to FtpClient being split into synd and async classes, see https://github.com/robinrodricks/FluentFTP/wiki/v40-Migration-Guide 
For now use the sync version of FtpClient (we can check if we need to switch to AsyncFtpClient) 
Update target framework to net6.0 and net6.0-windows
Removed packages-config since is no longer used (because we are using package references  https://learn.microsoft.com/en-us/nuget/reference/packages-config)
Update yml file to build with net6 on the CI

### DIFF
--- a/Activities/FTP/UiPath.FTP.Activities.Design/UiPath.FTP.Activities.Design.csproj
+++ b/Activities/FTP/UiPath.FTP.Activities.Design/UiPath.FTP.Activities.Design.csproj
@@ -1,11 +1,11 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk.WindowsDesktop" ToolsVersion="Current">
   <Import Project="..\FTP.build.props" />
   <PropertyGroup>
-    <TargetFrameworks>net461;net5.0-windows</TargetFrameworks>
+    <TargetFrameworks>net461;net6.0-windows</TargetFrameworks>
     <OutputPath>$(ProjectDir)..\..\Output\Activities\FTP\</OutputPath>
     <UseWPF>true</UseWPF>
   </PropertyGroup>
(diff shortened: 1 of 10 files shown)


Activities/FTP/UiPath.FTP/FtpObjectType.cs
```diff
@@ -1,6 +1,8 @@
 ﻿namespace UiPath.FTP
 {
-    public enum FtpObjectType
+    //Prefix the name with uipath due to conflict with FtpObjectType from FluentFTP
+    //see https://github.com/robinrodricks/FluentFTP/blob/8e7ba0ad892759c40b799748241bc4d20d8150f1/FluentFTP/Enums/FtpObjectType.cs#L7
+    public enum UiPathFtpObjectType
```
alexandru-petre: could we make it `internal`?

viogroza: yes, will do

viogroza: I double checked, we cannot make it internal since it's part of the the view model for the activity that enumerates the files
https://github.com/UiPath/Community.Activities/blob/4818fc6e3d22976fe6f1d4ffc72b03941e8b50c5/Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/EnumerateObjectsViewModel.cs#L42
I will prefix with the namespace name when necessary to avoid conflicts

alexandru-petre (approved):

PR #435, opened 2024-07-19
Database: Oracle Ref Cursor support (24.11)

viogroza:
Allow the type of the parameter to be changed
Change the type of the parameter to OracleDbType.RefCursor
Change the type of the parameter for a few know types

Key points:
Until now the type of the parameter was always string when passed to the underlying database service so oracle was unable to convert string to ref_cursor and that was causing the error
TBD what conversion we want to make (some of them will be breaking changes); most of the conversions work ok with string

### DIFF
--- a/Activities/Database/UiPath.Database/DatabaseConnection.cs
+++ b/Activities/Database/UiPath.Database/DatabaseConnection.cs
@@ -15,6 +15,7 @@ using UiPath.Database.BulkOps;
 using UiPath.Database.Properties;
 using UiPath.Data.ConnectionUI.Dialog.Workaround;
 using UiPath.Robot.Activities.Api;
+using Oracle.ManagedDataAccess.Types;
 
 namespace UiPath.Database
 {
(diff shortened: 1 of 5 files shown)


Activities/Database/UiPath.Database/DatabaseConnection.cs
```diff
@@ -551,5 +561,41 @@ private static ArgumentDirection WokflowParameterDirectionToDbParameter(Paramete
                     throw new ArgumentException(Resources.ParameterDirectionArgumentException);
             }
         }
+
+        private bool UpdateDbParamType(DbParameter dbParameter, ParameterInfo parameterInfo)
```
sorinconstantin: where is the return value of this method used?

viogroza: is not used at the moment..I will remove it
the idea is that in the future we can add the return value back
when changing the type of the parameter.. we want to change the most specific one then fallback to other 
some of the connections don't support changing the type of the param (eg. I tested with ODBC and Excel source)
Since there is a risk of breaking existing behavior I would change only what is strictly necessary (meaning for Oracle connection)

sorinconstantin (approved):

alexandru-petre (approved):

PR #444, opened 2024-10-14
FTP: Fixed viewmodel issues (72176, 72177, 72178)

alexandru-petre:
Enabled view models on Windows projects.
Customized the data source of the `SSLProtocols` property from the `UseFTPConnection` activity. (STUD-72176)
Fixed some resource strings.
Fixed the widgets of some boolean properties. (STUD-72177)
Cleaned the code of view models and fixed their appearance in StudioWeb. (STUD-72178)
Touched the solution files of several packages because the shared projects were not included in their solutions

https://uipath.atlassian.net/browse/STUD-72176
https://uipath.atlassian.net/browse/STUD-72177
https://uipath.atlassian.net/browse/STUD-72178


Activities/Shared/UiPath.Shared/SystemExtensions.cs
```diff
@@ -19,9 +19,9 @@ public static IComparer<T> ToComparer<T>(this Func<T, T, int> compare)
             return new Comparer<T>(compare);
         }
 
-        private class Comparer<T> : IComparer<T>
+        private sealed class Comparer<T> : IComparer<T>
```
viogroza, line 22: does the Extensions class need to be made internal also since is in a shared project?

Activities/Shared/UiPath.Shared.Activities/EnumExtensions.cs
```diff
@@ -0,0 +1,57 @@
+﻿using System;
+using System.CodeDom;
```
viogroza: unused namespaces

Activities/Shared/UiPath.Shared.Activities/EnumExtensions.cs
```diff
@@ -0,0 +1,57 @@
...
+using System.Linq;
+using System.Text;
+
+namespace UiPath.Shared.Activities
+{
+    internal static class EnumExtensions<T> where T : struct
+    {
+        public static bool EnumTryParseNoIntegers(string value, out T result)
+        {
+            if (int.TryParse(value, out _))
```
viogroza: why return default if we succeed to parse here?

Activities/Shared/UiPath.Shared.Activities/EnumExtensions.cs
alexandru-petre: fixed

Activities/Shared/UiPath.Shared.Activities/EnumExtensions.cs
alexandru-petre: this code was trying to eliminate integer values and allow only enum names, because Enum.TryParse will return true for non-existent integer values (it will return that value as if it's part of the enum)

anyway, the method was copied from system but it was useless so I've dropped it

Activities/Activities.Python.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.30621.155
```
viogroza, line 4: this change (and a few other) doesn't seem related to the jira tickets that are linked to this ticket

viogroza (approved):

Activities/Activities.Python.sln
alexandru-petre: source files from the shared projects were used but not included in the solution

PR #445, opened 2024-10-17
[WIP] Database: Fix/database query activities failing parameters conversion from DBNull (71685)

theloginerror:
Fixed ExecuteQuery and ExecuteNonQuery throwing an exception when trying to convert the parameter of type DBNull returned by the query:
`A value of type 'System.DBNull' cannot be set to the location with name 'Argument1' because it is a location of type 'System.String'`

https://uipath.atlassian.net/browse/STUD-71685

### DIFF
--- a/Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
+++ b/Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
@@ -105,7 +105,14 @@ namespace UiPath.Database.Activities
                     var currentParam = Parameters[param.Key];
                     if (currentParam.Direction == ArgumentDirection.Out || currentParam.Direction == ArgumentDirection.InOut)
                     {
-                        currentParam.Set(asyncCodeActivityContext, param.Value.Value);
+                        if (param.Value.Value != DBNull.Value)
+                        {
+                            currentParam.Set(asyncCodeActivityContext, param.Value.Value);
+                        }
+                        else
+                        {
+                            currentParam.Set(asyncCodeActivityContext, null);
+                        }
                     }
                 }
             };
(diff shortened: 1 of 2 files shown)


Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
```diff
@@ -105,7 +105,14 @@ protected async override Task<Action<AsyncCodeActivityContext>> ExecuteInternalA
...
                     if (currentParam.Direction == ArgumentDirection.Out || currentParam.Direction == ArgumentDirection.InOut)
                     {
-                        currentParam.Set(asyncCodeActivityContext, param.Value.Value);
+                        if (param.Value.Value != DBNull.Value)
+                        {
+                            currentParam.Set(asyncCodeActivityContext, param.Value.Value);
+                        }
+                        else
+                        {
+                            currentParam.Set(asyncCodeActivityContext, null);
```
viogroza, line 114: I would remove the condition on else
if there is nothing to set, I would not set null; (this is a business requirement)
for example if I have a in/out param with the initial value 100, then the procedures value for the param is DbNull.. do we want to set to zero the value in this case? or let it as it was? we can use default instead of null if needed


also null assignment wouldn't work if the type is a value type (eg. integer or bool);
eg.: 
object o = null;
bool v = (bool)o;

alexandru-petre:
Good fix — passing `DBNull.Value` directly to `Argument.Set()` would fail at runtime. Converting to `null` is the right direction.

One edge case to consider: **value types**. If an Out/InOut parameter is `InArgument<int>` and the procedure returns NULL, the current code calls `currentParam.Set(context, null)` which will throw — value types can't hold null.

The ADO.NET/Dapper/EF Core consensus for DBNull mapping is:
- **Reference types** (`string`) → `null`
- **Nullable value types** (`int?`) → `null`
- **Non-nullable value types** (`int`) → `default(T)` (0 for int, false for bool, etc.)

Suggested approach:

```csharp
if (param.Value.Value == DBNull.Value)
{
    var argType = currentParam.ArgumentType;
    if (!argType.IsValueType || Nullable.GetUnderlyingType(argType) != null)
    {
        currentParam.Set(asyncCodeActivityContext, null);
    }
    else
    {
        currentParam.Set(asyncCodeActivityContext, Activator.CreateInstance(argType));
    }
}
else
{
    currentParam.Set(asyncCodeActivityContext, param.Value.Value);
}
```

Same change applies to both `ExecuteNonQuery.cs` and `ExecuteQuery.cs` (identical pattern at the same location).

PR #451, opened 2024-11-11
Python: allow newer .net versions STUD-71044

alexandru-petre:
Set the RollForward=Major property in the project file.

https://uipath.atlassian.net/browse/STUD-71044

### DIFF
--- a/Activities/Python/UiPath.Python.Host/UiPath.Python.Host.csproj
+++ b/Activities/Python/UiPath.Python.Host/UiPath.Python.Host.csproj
@@ -2,6 +2,7 @@
   <Import Project="..\Python.build.props" />
   <PropertyGroup>
     <TargetFrameworks>net461;net6.0-windows;net6.0</TargetFrameworks>
+    <RollForward>Major</RollForward>
     <OutputPath>..\..\Output\Activities\Python\</OutputPath>
     <OutputType>Exe</OutputType>
 	<ExcludeFromSingleFile>true</ExcludeFromSingleFile>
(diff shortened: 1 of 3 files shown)


sorinconstantin:
My understanding is that RollForward helps the application to determine which installed version of, in this case framework, to use in case the one requested is not installed on the machine it runs on.
So, in the case described by the issue, how does it work? 🤔

sorinconstantin (approved):

viogroza (approved):

Activities/Python/UiPath.Python.Host/UiPath.Python.Host.csproj
```diff
@@ -2,6 +2,7 @@
   <Import Project="..\Python.build.props" />
   <PropertyGroup>
     <TargetFrameworks>net461;net6.0-windows;net6.0</TargetFrameworks>
+    <RollForward>Major</RollForward>
```
lucarachiteanu, line 5:
```suggestion
    <RollForward>LatestMajor</RollForward>
```
aren't we susceptible to this ?
> Minor version roll-forward has one side-effect that may affect end users. Consider the following scenario:
> 1. The application specifies that 5.0 is required.
> 2. When run, version 5.0.* isn't installed, however, 5.1.0 is. Version 5.1.0 will be used.
> 3. Later, the user installs 5.0.3 and runs the application again, 5.0.3 will now be used.
>
> It's possible that 5.0.3 and 5.1.0 behave differently, particularly for scenarios like serializing binary data.

and couldn't we avoid it by always using the latest? (Of course, a similar change in behavior can occur when rolling forward as well, but at least rolling forward is intended, while backwards is not)

alexandru-petre: I would leave Major because [it uses .net6 if it's available](https://learn.microsoft.com/en-gb/dotnet/core/versions/selection#values):

> **Major**: Roll-forward to the next available higher major version, and lowest minor version, if requested major version is missing. If the requested major version is present, then the Minor policy is used.
> **LatestMajor**: Roll-forward to highest major and highest minor version, _even if requested major is present_.

From a security / performance point of view, it would make sense indeed to use LatestMajor, but the further we are from the native framework, the higher the chances that the statement ".net X is backwards compatible with Y" comes with more fine-prints.

Thoughts?

viogroza (approved):

PR #456, opened 2024-12-16
FTP: Replace menu (25.2) [STUD-73198]

viogroza:
Make the toggle between string and secure string
At runtime use the actual mode instead of using password and if not set fallback to secure password

https://uipath.atlassian.net/browse/STUD-73198

### DIFF
--- a/Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs
+++ b/Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs
@@ -135,12 +135,12 @@ namespace UiPath.FTP.Activities.NetCore.ViewModels
             MenuActionsBuilder<PasswordInputMode>.WithValueProperty(PasswordInputModeSwitch)
               .AddMenuProperty(Password, PasswordInputMode.Password)
               .AddMenuProperty(SecurePassword, PasswordInputMode.SecurePassword)
-              .BuildAndInsertMenuActions();
+              .BuildAndInsertMenuActions(true);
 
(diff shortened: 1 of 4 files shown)


sorinconstantin (approved):

Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs
```diff
@@ -135,12 +135,12 @@ protected override void InitializeModel()
...
               .AddMenuProperty(Password, PasswordInputMode.Password)
               .AddMenuProperty(SecurePassword, PasswordInputMode.SecurePassword)
-              .BuildAndInsertMenuActions();
+              .BuildAndInsertMenuActions(true);
 
             MenuActionsBuilder<PasswordInputMode>.WithValueProperty(CertificatePasswordInputModeSwitch)
               .AddMenuProperty(ClientCertificatePassword, PasswordInputMode.Password)
               .AddMenuProperty(ClientCertificateSecurePassword, PasswordInputMode.SecurePassword)
-              .BuildAndInsertMenuActions();
+              .BuildAndInsertMenuActions(false);
```
alexandru-petre: I think that this one, although not mentioned explicitly in the ticket, suffers from the same lack of visibility and should be promoted to a main option as well. We should ask Maciej.

alexandru-petre (approved):

Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs
viogroza: I reached to Maciej and he confirmed that the changes should be made also for ClientCertificate
I will update this PR

PR #467, opened 2025-03-27
Database: Start transaction (25.4) [STUD-75123]

viogroza:
Fix StartTransaction not taking into account properly continueOnError 
Fix type cache issue for StartTransaction

https://uipath.atlassian.net/browse/STUD-74840
https://uipath.atlassian.net/browse/STUD-75123


Activities/Database/UiPath.Database.Activities/DatabaseTransaction.cs
```diff
@@ -126,13 +128,16 @@ private void OnCompletedCallback(NativeActivityContext context, ActivityInstance
...
+
+            HandleException(ex, continueOnError);
         }
 
         private void OnFaultedCallback(NativeActivityFaultContext faultContext, Exception exception, ActivityInstance source)
         {
             faultContext.CancelChildren();
             DatabaseConnection conn = DatabaseConnection.Get(faultContext);
             var continueOnError = ContinueOnError.Get(faultContext);
+            var primaryException = exception;
```
sorinconstantin, line 140: is there a situation when exception is null?

Activities/Database/UiPath.Database.Activities/DatabaseTransaction.cs
```diff
@@ -145,7 +150,9 @@ private void OnFaultedCallback(NativeActivityFaultContext faultContext, Exceptio
                 catch (Exception ex)
                 {
                     Trace.TraceError(ex.Message);
-                    HandleException(ex, continueOnError);
+                    //we should trace the original exception if present
+                    if (primaryException == null)
```
sorinconstantin, line 154: and regarding the previous comment, is this a plausible situation?

Activities/Database/UiPath.Database.Activities/DatabaseTransaction.cs
viogroza: from the tests I've done the exception is not null on faultedcallback
In theory it cannot be null
Both the case when exception is null and not null are handled properly

Activities/Database/UiPath.Database.Activities/DatabaseTransaction.cs
viogroza: Most likely the primaryException won't be null
There are a couple of scenarios:
1. primary exception not null -> the exception will be thrown (before it was ignored)
2. primary exception null and exception thrown when rolling back -> the new (rollback) exception is thrown
3. primary exception is null and rollback thrown no exception -> we will not throw

according to specification, the exception is most likely non-null, but it doesn't says that explicitly (I know for example that in a catch block the exception caught is always not null)
https://learn.microsoft.com/en-us/dotnet/api/system.activities.faultcallback?view=netframework-4.8.1

sorinconstantin (approved):

alexandru-petre (approved):

PR #473, opened 2025-05-28
Python: Cross-platform support (25.6) [STUD-75430]

viogroza:
Add view models for all activities
Remove Python.Runtime.Unix.dll (now use the same dll on all platforms) 
Make PythonHost cross-platform
Use dotnet to start the PythonHost on all platforms
Set default target to x64 for PythonScope
Fix build problem for all activities (changing files required rebuild) 
Remove UiPath.Activities.Design project (reference)
Remove some unused files

TO-DO (possibly address in a future PR)
- [ ] Clean generated nuget package (it seems that we are packing more files than are necessary)
- [ ] Enable/Add more unittests

https://uipath.atlassian.net/browse/STUD-75430


Activities/Directory.build.targets
```diff
@@ -76,4 +76,21 @@
 		<PackageReference Include="System.Threading.Tasks.Extensions" Version="4.5.4" />
 		<PackageReference Include="System.ValueTuple" Version="4.5.0" />
 	</ItemGroup>
+
+	<!-- For whatever reason, with "Fast UpToDate" VS option ON, changes are not detected properly 
+	Eg. obj/Debug/net6-0/refint/ .. is not copied to /obj/Debug/net6-0/ref even if there are changes
+	This causes package not to be generated even if there are changes (rebuild is necessary)
+	This is very annoying when developing -->
+	<Target Name="UpdateObjRefIfNeeded" AfterTargets="PostBuildEvent" Condition="$(ConfigurationName) == Debug">
```
alexandru-petre, line 84: According to multiple sources, including https://learn.microsoft.com/en-us/dotnet/core/compatibility/sdk/6.0/write-reference-assemblies-to-obj, the new default location for reference assemblies is obj/refint. The other folder, obj/ref, seems to be there for compatibility reasons (in case some scripts search for the old location).
As such, I think we should not copy from refint to ref, but rather delete the ref folder.

Activities/Python/UiPath.Python.Activities.ViewModels/ViewModels/PythonScopeViewModel.cs
```diff
@@ -0,0 +1,31 @@
...
+{
+    [ExcludeFromCodeCoverage]
+    class PythonScopeViewModel : DesignPropertiesViewModel
+    {
+        public PythonScopeViewModel(IDesignServices services) : base(services)
+        {
+            //Force dependency here for activities.api
+            //Otherwise this assembly will be ignored by studio
+            //see STUD-73622
+            var _ = services.GetService<IWorkflowDesignApi>();
```
alexandru-petre: Works fine on my computer, without this workaround.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -97,33 +95,30 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
+                        if (_version == Version.Python_310)
                         {
                             if (!string.IsNullOrEmpty(_libraryPath))
                                 _pyRuntime.PythonDLL = _libraryPath;
                         }
                         else
                             _pyEngine.PythonHome = _path;
 
-                        if (_version >= Version.Python_36 && _version <= Version.Python_39 || !_isWindows)
+                        if (_version >= Version.Python_36 && _version <= Version.Python_39)
```
alexandru-petre, line 121: Python 3.8 is passed the end of life, I think we should remove support for it and anything older. I've created a separate ticket for this task.

Activities/Python/UiPath.Python.Activities.ViewModels/ViewModels/PythonScopeViewModel.cs
viogroza: ~~I will double check~~
This workaround is not needed here because a reference is forced here https://github.com/UiPath/Community.Activities/blob/b7790ff00781870d7619fc669ecdd22c0e1045ad/Activities/Python/UiPath.Python.Activities.ViewModels/ViewModels/GetObjectViewModel.cs#L29

Without the reference (without both workaround and the one in GetObjectViewModel) it doesn't work. The view models are not loaded. This is a know issue https://uipath.atlassian.net/browse/STUD-73622

I tested and it locally (after removing the reference) and it doesn't work
image

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Thanks

Activities/Directory.build.targets
viogroza: ~~I will double check~~
I tried with a sample project.
I checked and **deleting the contains from ref folder doesn't** work since is needed for building dependencies
If I delete the file I have this error
3>CSC : error CS0006: Metadata file '..\Module2\obj\Debug\net6.0\ref\Module2.dll' could not be found

(When fast build is enabled https://devblogs.microsoft.com/visualstudio/vs-toolbox-accelerate-your-builds-of-sdk-style-net-projects/) 
dll is not copied in the ref folder _if public api_ has not changed (what the dll exports)
However, the file in ref is checked when deciding if the dependencies are changed
In most cases, this is not problem, but for custom builds is (eg. in our case we are creating the package on build for debug config)

For the dependency, when CopyRefAssembly is executed
_Reference assembly "..Module2\obj\Debug\net6.0\ref\Module2.dll" already has latest information. Leaving it untouched._
Then when checking for the file (and seeing dependency as unchanged)
No inputs are newer than earliest output '..CompileTest\obj\Debug\net6.0-windows\CompileTest.pdb' (2025-05-27 19:46:10.308). Newest input is '..Module2\obj\Debug\net6.0\ref\Module2.dll' (2025-05-27 19:46:09.319). (CompileTest)

alexandru-petre (approved):

PR #486, opened 2025-07-30
FTP: EnumerateObjects filter (25.8) [STUD-76148]

viogroza:
Move package versions to props and targets files
Add specific Directory.build.props
Add specific Directory.build.targets
Add missing Continue on error property for all FTP activities (it was hidden by mistake when moving to viewmodels) Add filter property for EnumerateObjects
Cleanup package project
Add unittests

https://uipath.atlassian.net/browse/STUD-76148

image

image

### DIFF
--- a/Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/EnumerateObjectsViewModel.cs
+++ b/Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/EnumerateObjectsViewModel.cs
@@ -1,12 +1,15 @@
-﻿using System.Activities.DesignViewModels;
+﻿using System;
+using System.Activities.DesignViewModels;
 using System.Activities.ViewModels;
-using System.Threading.Tasks;
-using UiPath.FTP.Activities.NetCore.ViewModels;
-
+using System.Collections.Generic;
+using System.Linq;
+using UiPath.FTP.Activities.Properties;
+using UiPath.Shared;
+using UiPath.Shared.Activities;
 
 namespace UiPath.FTP.Activities.NetCore.ViewModels
 {
-    public partial class EnumerateObjectsViewModel : DesignPropertiesViewModel
+    internal class EnumerateObjectsViewModel : BaseFtpViewModel
     {
         /// <summary>
         /// Basic constructor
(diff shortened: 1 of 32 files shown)


Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -229,13 +234,14 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
                 }
             }
 
+            IFtpSession ftpSession = _setFtpSession;
             if (UseSftp)
             {
-                ftpSession = new SftpSession(ftpConfiguration);
+                ftpSession ??= new SftpSession(ftpConfiguration);
```
sorinconstantin, line 240: shouldn't this be `_ftpSession`

Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/EnumerateObjectsViewModel.cs
```diff
@@ -40,9 +51,82 @@ protected override void InitializeModel()
...
+            return string.Join(" | ", labels);
+        }
+
+        /// <summary>
+        /// Returns the localized name of an enum flag (i.e. a single enum value, not a combination).
+        ///  Localized names of obsolete values are marked appropriately.
+        /// </summary>
+        /// <param name="flag"></param>
+        /// <returns>The localized name</returns>
+        private static string GetFilteObjectTypeFlagLabel(FtpFilterObjectType flag)
```
sorinconstantin: nit: typo 
```suggestion
        private static string GetFilterObjectTypeFlagLabel(FtpFilterObjectType flag)
```

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
viogroza: It's set below when we schedule the body activity
_This is the difference if in a foreach scope or something similar: currently the connection is created on every iteration.. If I change as you suggested it will reuse always the same connection._
I hesitated to change to _ftpSession since this can get us in trouble when running in foreach. I mean, I didn't wanted to introduce a potential breaking change
We could argue if this is the correct approach or not.. It's just the old behavior and I didn't wanted to change it
Please note that _setFtpSession is set only by unittests

Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/EnumerateObjectsViewModel.cs
viogroza: will correct it, ty

sorinconstantin (approved):

PR #501, opened 2025-11-17
Refactor telemetry handling and resource cleanup

AlexMarinescuUiPath:
Refactored telemetry operation handling in `ExecuteAsync`, `OnCompleted`, and `OnFaulted` methods to improve consistency and maintainability. Introduced conditional compilation (`#if ENABLE_DEFAULT_TELEMETRY`) for telemetry operations, reducing redundant code. Ensured proper sequencing of telemetry operations and resource cleanup (`_ftpSession.Close()` and `_ftpSession.Dispose()`) in all relevant methods.

### DIFF
--- a/Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
+++ b/Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
@@ -182,6 +182,11 @@ namespace UiPath.FTP.Activities
 
         protected override async Task<Action<NativeActivityContext>> ExecuteAsync(NativeActivityContext context, CancellationToken cancellationToken)
         {
+            ITelemetryOperationWrapper telemetryOperation = null;
+#if ENABLE_DEFAULT_TELEMETRY
+            telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
+#endif
+
             try
             {
                 string passwordValue = Password.Get(context);
@@ -260,11 +265,7 @@ namespace UiPath.FTP.Activities
             }
             catch (Exception ex)
             {
-#if ENABLE_DEFAULT_TELEMETRY
-                ITelemetryOperationWrapper telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
                 telemetryOperation?.SendWithException(ex);
-#endif
-
                 throw;
             }
 
--- a/Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
+++ b/Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
@@ -182,11 +182,6 @@ namespace UiPath.FTP.Activities
 
         protected override async Task<Action<NativeActivityContext>> ExecuteAsync(NativeActivityContext context, CancellationToken cancellationToken)
         {
-            ITelemetryOperationWrapper telemetryOperation = null;
-
-#if ENABLE_DEFAULT_TELEMETRY
-            telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
-#endif
             try
             {
                 string passwordValue = Password.Get(context);
@@ -260,14 +255,16 @@ namespace UiPath.FTP.Activities
                         _ftpSession = ftpSession;
                         nativeActivityContext.ScheduleAction(Body, ftpSession, OnCompleted, OnFaulted);
                     }
-                    else
-                        telemetryOperation?.Send();
                 });
                 return result;
             }
             catch (Exception ex)
             {
+#if ENABLE_DEFAULT_TELEMETRY
+                ITelemetryOperationWrapper telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
                 telemetryOperation?.SendWithException(ex);
+#endif
+
                 throw;
             }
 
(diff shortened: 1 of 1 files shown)


Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -260,46 +255,44 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
...
-                        telemetryOperation?.Send();
                 });
                 return result;
             }
             catch (Exception ex)
             {
+#if ENABLE_DEFAULT_TELEMETRY
+                ITelemetryOperationWrapper telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
                 telemetryOperation?.SendWithException(ex);
+#endif
```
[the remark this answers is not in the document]
AlexMarinescuUiPath: did so because it is only used in the catch block

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -182,11 +182,6 @@ protected override void CacheMetadata(NativeActivityMetadata metadata)
...
         protected override async Task<Action<NativeActivityContext>> ExecuteAsync(NativeActivityContext context, CancellationToken cancellationToken)
         {
-            ITelemetryOperationWrapper telemetryOperation = null;
-
-#if ENABLE_DEFAULT_TELEMETRY
-            telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
-#endif
             try
             {
                 string passwordValue = Password.Get(context);
```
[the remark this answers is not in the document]
AlexMarinescuUiPath: no, other scope type activities in this package don't do so; they only send Success OnCompleted

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -260,46 +255,44 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
...
+#if ENABLE_DEFAULT_TELEMETRY
+            ITelemetryOperationWrapper telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
+            telemetryOperation?.Send();
+#endif
+
             if (_ftpSession == null)
                 throw new InvalidOperationException(Resources.FTPSessionNotFoundException);
 
             _ftpSession.Close();
             _ftpSession.Dispose();
```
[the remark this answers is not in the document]
AlexMarinescuUiPath: ideally, the logic should have been in a try-catch but that would imply handling success and error cases
we settled to send telemetry first, with whatever context, then handle the rest

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -260,46 +255,44 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
...
                     }
-                    else
-                        telemetryOperation?.Send();
                 });
                 return result;
             }
             catch (Exception ex)
             {
+#if ENABLE_DEFAULT_TELEMETRY
+                ITelemetryOperationWrapper telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, context);
```
lucarachiteanu: I am not sure this is safe, the context is either discarded or in a different state after the first await. I think CreateExecutionOperation needs to be created at the start of the telemetry operation. Moreover, shouldn't we also send on success?

sorinconstantin (approved):

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -260,46 +255,44 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
...
                         nativeActivityContext.ScheduleAction(Body, ftpSession, OnCompleted, OnFaulted);
                     }
-                    else
-                        telemetryOperation?.Send();
                 });
                 return result;
             }
             catch (Exception ex)
             {
+#if ENABLE_DEFAULT_TELEMETRY
```
lucarachiteanu: Same as the comment in another PR, I do not see this directive being used anywhere. If this should be optional, I think a no-op implementation type of thing is better, moreover, we could allow this to be in the control of the host, such as the Robot.

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
AlexMarinescuUiPath: I moved it in the catch statement, thinking that we only use it here but I think you are right, the context may not be the same
will revert this

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
AlexMarinescuUiPath: using this directive seems to be the pattern; do we want to talk about this and make a consistent change everywhere?

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
AlexMarinescuUiPath: also, in scope like activities, we only track exceptions and send successful telemetry only OnCompleted

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -260,46 +255,44 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
...
-
 #if ENABLE_DEFAULT_TELEMETRY
-            telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, faultContext);
-#endif
+            ITelemetryOperationWrapper telemetryOperation = RuntimeTelemetryService.CreateExecutionOperation(this, faultContext);
             telemetryOperation?.SendWithException(propagatedException);
+#endif
+
+            _ftpSession?.Close();
+            _ftpSession?.Dispose();
```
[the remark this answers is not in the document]
AlexMarinescuUiPath: intended

PR #514, opened 2026-02-25
feat(cryptography): add PGP support for encryption, decryption, signing, and verification

adrianignat13:
## Summary

The Cryptography activity pack only supported symmetric algorithms (AES, DES, RC2, Rijndael). Users needed PGP (asymmetric, key-based) cryptography for encrypting/decrypting files and text, signing files, clear-signing files, verifying signatures, and generating key pairs — common requirements in enterprise automation workflows.

## Changes

**Extended existing activities** (EncryptFile, DecryptFile, EncryptText, DecryptText) with a PGP algorithm option that dynamically shows/hides public key, private key, and passphrase inputs based on the selected algorithm.

**Added 4 new activities:**
- PgpSignFile — binary signature
- PgpClearSignFile — clear-text signature
- PgpVerify — unified signature, clear-signature, and public-key verification
- PgpGenerateKeyPair — key pair generation

**Added helpers:**
- PgpStreamHelper — key file validation and stream lifecycle management
- PgpFileSignHelper — shared sign/clear-sign logic

**Reduced ViewModel duplication** with shared base classes: EncryptCryptoViewModelBase, DecryptCryptoViewModelBase, and PgpSignViewModelBase.

**Added ViewModels** for all new and modified activities with full metadata registration and SVG icons.

**Added tests:** PgpTests (activity-level encrypt/decrypt/sign/verify round-trips) and PgpStandaloneTests (helper-level unit tests), with shared setup extracted into PgpTestBase.

**Renamed** SymmetricAlgorithms.cs → EncryptionAlgorithm.cs (file rename only — the enum was already named EncryptionAlgorithm).

**Bug fixes and cleanup:**
- Fixed pre-existing bug in DecryptText.cs where ContinueOnError description pointed to Result_Description
- Added argument validation: PGP encrypt/decrypt now throws ArgumentException when sign/verify is requested but required inputs (private key, passphrase, public key) are missing, instead of silently falling back to unsigned/unverified operation
- Refactored TranslatePgpException to return Exception instead of void, eliminating unreachable `return null` workarounds and preserving original stack traces via ExceptionDispatchInfo
- Added NonClosingStreamWrapper in PgpVerifyPublicKey to prevent BouncyCastle's decoder from closing the caller-provided stream
- Added LocalizedDisplayName/LocalizedDescription attributes to all PGP properties for consistency with existing activity properties
- Removed CopyLocalLockFileAssemblies from the library project; PgpCore dependency moved to the packaging project
- Removed static IsRequired from PgpVerify metadata for InputFilePath/PublicKeyFilePath — ViewModel handles this dynamically per verification mode
- Cleaned up SonarQube findings: removed unnecessary casts, redundant null-conditional operators, unused imports, and replaced Path.GetTempFileName with Path.GetRandomFileName in tests
- Removed orphaned resource strings from consolidated verify activities

Uses [PgpCore](https://github.com/mattosaurus/PgpCore) NuGet package (v6.5.3) as the underlying PGP implementation.

43 files changed — 4,655 insertions, 936 deletions.

## Test plan
- [x] Build: 0 errors (warnings are all pre-existing)
- [x] Tests: 73/73 passing


sorinconstantin (approved):
nit: LocalizedDescriptionAttribute usage on properties is redundant when view models exist. LocalizedDisplayName should also be used only for properties marked as Required, so the validation message is pointing to the correct display name of a localized property

Activities/Cryptography/UiPath.Cryptography.Activities/UiPath.Cryptography.Activities.csproj
```diff
@@ -26,6 +26,10 @@
  <EmbeddedResource Include="Resources\Icons\Hash_file_with_key.svg" />
  <EmbeddedResource Include="Resources\Icons\Hash_text.svg" />
  <EmbeddedResource Include="Resources\Icons\Hash_text_with_key.svg" />
+ <EmbeddedResource Include="Resources\Icons\PGP_generate_keys.svg" />
+ <EmbeddedResource Include="Resources\Icons\PGP_sign_file.svg" />
+ <EmbeddedResource Include="Resources\Icons\PGP_clear_sign_file.svg" />
+ <EmbeddedResource Include="Resources\Icons\PGP_verify.svg" />
```
lucarachiteanu, line 32: nit: we can do this with a wild card

Activities/Cryptography/UiPath.Cryptography/EncryptionAlgorithm.cs
```diff
@@ -1,9 +1,9 @@
-﻿using System;
+using System;
 using UiPath.Cryptography.Properties;
 
 namespace UiPath.Cryptography
 {
-    public enum SymmetricAlgorithms
+    public enum EncryptionAlgorithm
```
lucarachiteanu, line 6: Isn't this a breaking change?

adrianignat13: tl;dr No

explanation: the Algorithm property only accepts a value from the drop-down and that is what we store in the xaml. It could break if the user insists on having a variable/argument of that type and then on upgrade the type is no longer there, but having such a variable makes no sense.

PR #548, opened 2026-05-09
chore(l10n): sync from Localization

andrei-stoian:
The build that generated this PR: https://uipath.visualstudio.com/DevOps/_build/results?buildId=11974538


alexandru-petre:
Replaced by a newer PR.

PR #549, opened 2026-05-11
Python: No installations message [STUD-77361]

viogroza:
Show -No Python installations were detected- when no installations are found

https://uipath.atlassian.net/browse/STUD-77361

image


alexandru-petre:
@adrianignat13 has been assigned to this PR.

alexandru-petre:
@sorinconstantin has been assigned to this PR.

sorinconstantin (changes requested):
High: `Activities/Python/UiPath.Python.Activities.ViewModels/ViewModels/PythonScopeViewModel.cs:94` has the
  condition reversed. `if (GetInstalledPythonVersions().Any())` enters the “No Python installations were detected” path
  when installations do exist, making the dropdown read-only and hiding valid choices. The check should be `!
  GetInstalledPythonVersions().Any()` or use a local `installedVersions` list and branch on `Count == 0`.

viogroza:
> `installedVersions`

That is correct.. I initially let this condition as it is for testing purposes (since I had python installations on the machine)
I will correct it

sorinconstantin (approved):
Only a small nit remains: GetInstalledPythonVersions() is invoked four times in this method. It's backed by a Lazy<IReadOnlyList<…>> so the cost is trivial — purely a readability thought; a local var installed =
  GetInstalledPythonVersions(); would tidy it.

PR #555, opened 2026-05-19
Python: Pipe issues [STUD-80063] [STUD-80064] [STUD-80065]

viogroza:
Add ScriptDataSizeLimitMB and LogTraces properties to PythonScope Update viewmodels to reflect the new properties
Leave pipe communication exceptions to propage
Add handlers for OutputDataReceived and ErrorDataReceived 
Log python traces when LogTraces is active
Throw exception when the size of the parameters is too big
Add execution/serialization support for collection params (list or dictionaries)
Prevent exceptions being swallowed
Add better descriptions for exceptions

https://uipath.atlassian.net/browse/STUD-80063
https://uipath.atlassian.net/browse/STUD-80064
https://uipath.atlassian.net/browse/STUD-80065

TODO:

- [ ] Add unittests (I added this separate task for tracking since involves some refactoring https://uipath.atlassian.net/browse/STUD-80106) - too risky for this release


Activities/Python/UiPath.Python.Host.Shared/Program.cs
```diff
@@ -16,7 +17,7 @@ private static void Main()
             AppDomain.CurrentDomain.ProcessExit += Application_ApplicationExit;
 
             _service = new PythonService();
-            _service.RunServer();
+            _ = _service.RunServer(_cts.Token);
```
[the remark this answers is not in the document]
viogroza: will address, ty

Activities/Python/UiPath.Python/EngineProvider.cs
```diff
@@ -24,7 +26,7 @@ public static class EngineProvider
         private static object _lock = new object();
         private static Dictionary<Version, IEngine> _cache = new Dictionary<Version, IEngine>();
 
-        public static IEngine Get(Version version, string path, string libraryPath, bool inProcess = true, TargetPlatform target = TargetPlatform.x86, bool visible = false)
+        public static IEngine Get(Version version, string path, string libraryPath, bool inProcess = true, TargetPlatform target = TargetPlatform.x86, bool visible = false, bool logTrace = false, int payloadThresholdMB = DefaultPayloadThresholdMB)
         {
```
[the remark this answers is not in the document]
viogroza: it's ok to change here

Activities/Python/UiPath.Python.Host.Shared/PythonService.cs
```diff
@@ -55,7 +73,7 @@ internal async void RunServer()
                 {
                     try
                     {
-                        var request = PythonRequest.Deserialize(await streamReader.ReadLineAsync());
+                        var request = PythonRequest.Deserialize(await streamReader.ReadLineAsync().WaitAsync(ct));
```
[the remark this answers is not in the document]
viogroza: will fix, ty

Activities/Python/UiPath.Python/Service/Argument.cs
```diff
@@ -16,55 +18,168 @@ public class Argument
...
+            }
+            if (obj is IDictionary dict && TryGetDictionaryKeyValueTypes(type, out _, out _))
+            {
+                return WrapDictionary(dict);
+            }
+            if (obj is IEnumerable enumerable && !(obj is string))
+            {
+                return WrapEnumerable(enumerable);
+            }
+            return obj;
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Python/UiPath.Python/Properties/UiPath.Python.Designer.cs
```diff
@@ -140,5 +185,14 @@ public static string UnsupportedVersionException {
...
             }
         }
+
+        /// <summary>
+        ///   Looks up a localized string similar to The serialized Python argument exceeds the allowed payload threshold.
+        /// </summary>
+        public static string PayloadThresholdExceeded {
+            get {
+                return ResourceManager.GetString("PayloadThresholdExceeded", resourceCulture);
+            }
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Python/UiPath.Python.Activities/Properties/UiPath.Python.Activities.Designer.cs
```diff
@@ -276,6 +285,24 @@ public static string LoadScriptNameDisplayName {
...
+        }
+        
+        /// <summary>
+        ///   Looks up a localized string similar to Log Python Traces.
+        /// </summary>
+        public static string LogTracesDisplayName {
+            get {
+                return ResourceManager.GetString("LogTracesDisplayName", resourceCulture);
+            }
+        }
```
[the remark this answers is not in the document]
viogroza: solved in latest commit (I'm not sure why it doesn't show it as outdated - if you view the file you notice that the comment is: Log Python Output to File (Diagnostic))
see https://github.com/UiPath/Community.Activities/blob/83dd5e51910bbc9d62bcc7b14e3728a0f63e25e1/Activities/Python/UiPath.Python.Activities/Properties/UiPath.Python.Activities.Designer.cs#L300

viogroza:
rebased

Activities/Python/UiPath.Python/EngineProvider.cs
```diff
@@ -24,7 +27,7 @@ public static class EngineProvider
         private static object _lock = new object();
         private static Dictionary<Version, IEngine> _cache = new Dictionary<Version, IEngine>();
 
-        public static IEngine Get(Version version, string path, string libraryPath, bool inProcess = true, TargetPlatform target = TargetPlatform.x86, bool visible = false)
+        public static IEngine Get(Version version, string path, string libraryPath, bool inProcess = true, TargetPlatform target = TargetPlatform.x86, bool visible = false, bool logTrace = false, int payloadThresholdMB = DefaultPayloadThresholdMB)
```
alexandru-petre, line 30: **[P2-Compatibility] MEDIUM — Binary-incompat change on public API**

Signature changed from `Get(Version, string, string, bool, TargetPlatform, bool)` to `Get(Version, string, string, bool, TargetPlatform, bool, bool, int)`. Defaults make it source-compatible, but optional parameters are baked at the **call site** — any assembly compiled against the previous `UiPath.Python.dll` will throw `MissingMethodException` at load.

`UiPath.Python` is shipped as a public NuGet. External consumers (or downstream packages embedded against the previous version) will not bind to the new method.

**Suggested fix** — add an overload preserving the original signature:

```csharp
// Keep old signature for binary compat
public static IEngine Get(Version version, string path, string libraryPath,
    bool inProcess = true, TargetPlatform target = TargetPlatform.x86, bool visible = false)
    => Get(version, path, libraryPath, inProcess, target, visible,
           logTrace: false, payloadThresholdMB: DefaultPayloadThresholdMB);

public static IEngine Get(Version version, string path, string libraryPath,
    bool inProcess, TargetPlatform target, bool visible, bool logTrace, int payloadThresholdMB)
{ ... }
```

Activities/Python/UiPath.Python.Activities/PythonScope.cs
```diff
@@ -134,21 +149,33 @@ protected override async Task<Action<NativeActivityContext>> ExecuteAsync(Native
...
+                    if (autodetected != Version)
                         throw new InvalidOperationException(string.Format(Resources.InvalidVersionException, Version.ToFriendlyString(), autodetected.ToFriendlyString()));
                 }
 
+                // Python 3.10 requires an explicit library file (python**.dll on Windows,
+                // libpython*.so on Linux).
+                if (effectiveVersion == Version.Python_310 && (libraryPath.IsNullOrEmpty() || !File.Exists(libraryPath)))
+                    throw new FileNotFoundException(string.Format(Resources.InvalidLibraryPathException, libraryPath));
+
+                _pythonEngine = EngineProvider.Get(Version, path, libraryPath, !Isolated, TargetPlatform, ShowConsole, LogTraces, payloadThresholdMB);
```
alexandru-petre: **[P4-Performance] MEDIUM — `Autodetect` invoked twice when `Version=Auto`**

`EngineProvider.ResolveEffectiveVersion` (called above at line 159) runs `Autodetect`. Then `EngineProvider.Get(Version, …)` on this line *also* runs `Autodetect` internally when `Version=Auto` (because `Auto.IsValid()` is false). Each `Autodetect` spawns up to 6 Python child processes (`python --version` for each candidate path/exe combo).

Doubles cold-start cost of every PythonScope when the user hasn't pinned a version — several hundred ms of avoidable latency on slow disk or AV-scanned environments.

**Suggested fix** — pass the already-resolved version into `EngineProvider.Get` so it skips its internal autodetect:

```csharp
_pythonEngine = EngineProvider.Get(effectiveVersion, path, libraryPath, !Isolated,
                                   TargetPlatform, ShowConsole, LogTraces, payloadThresholdMB);
```

or cache the autodetect result in `EngineProvider` keyed by `path`.

---

**[P1-Bugs] LOW — Engine instance leak window between `Get` and `Initialize` (pre-existing, extended)**

The x86 check immediately after this line throws *after* `EngineProvider.Get(...)` has constructed the engine. For `OutOfProcessEngine` this is currently a no-op leak (the constructor doesn't start a process — `Initialize` does), but the order is fragile: any future change that does work in the engine constructor would leak a process on the `x86` throw.

**Suggested fix** — move the `Python_310 + x86` check above the `Get` call (it no longer needs `_pythonEngine.Version` because `effectiveVersion` is already resolved):

```csharp
if (effectiveVersion == Version.Python_310 && TargetPlatform == TargetPlatform.x86)
    throw new InvalidOperationException(Resources.ValidationErrorPlatformUnsupported);

_pythonEngine = EngineProvider.Get(effectiveVersion, path, libraryPath, !Isolated,
                                   TargetPlatform, ShowConsole, LogTraces, payloadThresholdMB);
```

Activities/Python/UiPath.Python.Activities.API/Services/PythonService.cs
```diff
@@ -55,7 +59,31 @@ public async Task<IPythonScopeHandle> UsePythonScope(PythonScopeOptions options,
...
+            // reported with clear errors instead of opaque native-load failures.
+            var effectiveVersion = EngineProvider.ResolveEffectiveVersion(options.Version, path, out var autodetected);
+
+            if (options.Version != Version.Auto && autodetected != Version.Auto)
+            {
+                if (!VersionExtensions.GetSupportedVersions().Contains(autodetected))
+                    throw new InvalidOperationException($"Python version '{autodetected}' is not supported.");
+                if (autodetected != options.Version)
+                    throw new InvalidOperationException(
+                        $"Python version mismatch: expected '{options.Version.ToFriendlyString()}' but found '{autodetected.ToFriendlyString()}' at path '{path}'.");
```
alexandru-petre: **[P3-Maintainability] MEDIUM — New error messages bypass the localization Resources**

Three new throws in this method use hard-coded English string interpolation:
- L42 `$"Python version '{options.Version}' is not supported."`
- L70 `$"Python version '{autodetected}' is not supported."`
- L72-73 `$"Python version mismatch: expected '{options.Version.ToFriendlyString()}' but found '{autodetected.ToFriendlyString()}' at path '{path}'."`
- L81-82 `$"The specified Python library path is not valid (a file path to python**.dll / libpython*.so is required for Python 3.10): {options.LibraryPath}"`

The equivalent strings already exist as resx entries: `ValidationErrorVersionUnsupported`, `InvalidVersionException`, and the new `InvalidLibraryPathException` added in this same PR. `PythonScope.cs` uses `Resources.*` for the exact same conditions — the API path now diverges, so error messages won't be translated and message text for the same condition will differ between XAML and Coded workflows.

**Suggested fix**:

```csharp
if (!VersionExtensions.GetSupportedVersions().Contains(autodetected))
    throw new InvalidOperationException(Resources.ValidationErrorVersionUnsupported);
if (autodetected != options.Version)
    throw new InvalidOperationException(
        string.Format(Resources.InvalidVersionException,
                      options.Version.ToFriendlyString(), autodetected.ToFriendlyString()));
…
throw new FileNotFoundException(string.Format(Resources.InvalidLibraryPathException, options.LibraryPath));
```

Activities/Shared/UiPath.Shared.Service/Client/HostWrapper.cs
```diff
@@ -72,16 +147,94 @@ internal bool GetHostProcessId(out int processId)
...
-                    string error = readerError.ReadToEnd();
-                    throw new Exception($"Host process has exited!\n output: {output} \n error: {error} \n");
+                    Trace.TraceWarning($"Disabling Python host diagnostic log after write failure: {ex.Message}");
+                    try { _logFile.Dispose(); } catch { /* best-effort */ }
+                    _logFile = null;
                 }
             }
         }
+
+        private StreamWriter OpenDiagnosticLogFile()
```
alexandru-petre, line 211: **[P5-Security] MEDIUM — `LogTraces` writes raw stdout/stderr to disk; no size cap, no rotation, no retention**

When enabled, every line of `print(...)` / stderr from the user's Python script is persisted to `%LOCALAPPDATA%\UiPath\Logs\python\python-host-<ts>-<pid>.log`. A script that does `print(api_key)` or `print(customer_record)` writes that to disk where it lives indefinitely.

Default is `false` and the description warns about production use, but operators who flip it on once for diagnosis can easily forget it. The diagnostic file is also not joined with the existing Robot/Studio log lifecycle, so retention policies for those logs don't apply here.

**Suggested mitigations**:
1. Hard size cap per file (e.g. 50 MB → roll over / stop)
2. Max number of files retained (delete oldest beyond N)
3. Clear marker line at the top of each file describing the risk
4. Optionally: validate the path resolves under the `LocalApplicationData` root before writing (defense against `LogsFolderOverride` misuse from tests bleeding into prod)

Activities/Shared/UiPath.Shared.Service/Enums.cs
```diff
@@ -18,14 +18,13 @@ internal enum RequestType
 
     internal enum ResultState
     {
+        Successful,
```
alexandru-petre, line 21: **[P2-Compatibility] LOW — `ResultState` lost six values**

Removed: `IllegalArguments`, `UnknownException`, `ScriptNotLoaded`, `ScriptNotFound`, `ScriptAlreadyLoaded`, `FieldNotFound`. The enum is serialized **by name** (see `PythonResponse.ResultStateString`), so reordering is wire-safe — but removed names cause `Enum.Parse` to throw on receipt.

The Activities pack ships the host and client together, so mixed-version operation is unlikely. The removed values are not produced by this PR's host either (the pre-PR host only ever sent `InstantiationException` and `Successful`), so on the wire this is benign today. Worth confirming there isn't a packaged scenario (e.g. cached `.dll` left in a Robot install directory after upgrade) that could ship a mismatch.

**Suggested fix** — either keep the removed values marked `[Obsolete]` for one release cycle to absorb stale clients, or explicitly call out in the PR description that mixed host/client versions are unsupported and the package upgrade replaces both.

Activities/Shared/UiPath.Shared.Service/Client/Controller.cs
```diff
@@ -41,30 +41,42 @@ internal void ForceStop()
...
-                hostFullPath = Path.Combine(folder, hostFile);
-            }
+            var (folder, hostFullPath) = ResolveHostFullPath(hostFile);
 
             if (!File.Exists(hostFullPath))
                 throw new Exception($"Process path not found: {hostFullPath}");
 
+            PythonWrapper.Proc = Process.Start(CreateProcessStartInfo(hostFullPath, folder, isExeMode));
+
+            // Note: event handlers must be subscribed before BeginOutputReadLine/BeginErrorReadLine,
```
alexandru-petre: **[P3-Maintainability] LOW — Comment is misleading**

The comment says "event handlers must be subscribed before BeginOutputReadLine/BeginErrorReadLine, but both require the process to already be started — subscriptions cannot move before Process.Start." The first half is correct. The second half is misleading: `OutputDataReceived += …` can be subscribed before `Process.Start` just fine — the documented constraint is "before `BeginOutputReadLine`," not "after `Process.Start`."

A future reader may move the subscriptions before `Process.Start` thinking the comment is wrong, then trip on something else. Or — more realistically — the comment hides that the current code has no actual ordering hazard.

**Suggested fix** — trim to the load-bearing fact, e.g.:

```csharp
// Subscribe before BeginOutputReadLine — these handlers drive the diagnostic log
// and the post-mortem buffer surfaced by ThrowIfProcessHasExited.
```

Activities/Python/UiPath.Python/Properties/UiPath.Python.Designer.cs
```diff
@@ -29,7 +29,7 @@ public class UiPath_Python {
         private static global::System.Globalization.CultureInfo resourceCulture;
         
         [global::System.Diagnostics.CodeAnalysis.SuppressMessageAttribute("Microsoft.Performance", "CA1811:AvoidUncalledPrivateCode")]
-        public UiPath_Python() {
+        internal UiPath_Python() {
```
alexandru-petre, line 32: **[P3-Maintainability] LOW — Auto-generated constructor visibility changed `public` → `internal`**

This is an auto-generated file (`<auto-generated>` header); the change is presumably the side effect of toggling the resx CustomTool access modifier in VS.

Almost certainly nobody calls `new UiPath_Python()` externally — the class only exposes static resource accessors. But it is a binary-incompatible change on a public-API surface in a public NuGet, and it's a drift from the existing generator default that isn't explained anywhere in the PR.

**Suggested action** — confirm whether this was intentional:
- If yes, mention it in the PR description.
- If it's an unintended VS regen artifact, revert the access change in this generated file (or, better, change the resx's Access Modifier setting back to `Public` in VS so the next regen produces `public` again).

Activities/Python/UiPath.Python/Service/PythonProxy.cs
```diff
@@ -25,12 +25,17 @@ internal class PythonProxy : IPythonService
...
-        internal PythonProxy(HostWrapper hostWrapper, double timeout, CancellationToken cancellationToken)
+        private const long MinPayloadThresholdBytes = 1L * 1024 * 1024; // 1 MB
+
+        internal PythonProxy(HostWrapper hostWrapper, double timeout, CancellationToken cancellationToken, int payloadThresholdMB)
         {
             PythonHostWrapper = hostWrapper;
             Timeout = timeout;
             Token = cancellationToken;
+            long requestedBytes = (long)payloadThresholdMB * 1024 * 1024;
+            PayloadThresholdBytes = Math.Max(requestedBytes, MinPayloadThresholdBytes);
```
alexandru-petre, line 38: **[P4-Performance] LOW — `Math.Max` silently overrides invalid values**

After all the upstream validation (in `PythonScope.CacheMetadata`, `PythonScope.ExecuteAsync`, and `PythonService.UsePythonScope`), every reachable caller has already validated `>= MinPayloadThresholdMB`. The `Math.Max` here is dead defensive code that would *hide* a future bug where a caller forgot the validation.

Defensive coding that silently corrects vs. throws on contract violation is the same anti-pattern flagged elsewhere in the codebase — the only thing worse than failing is silently succeeding with wrong values.

**Suggested fix** — either remove the clamp (callers validate), or replace it with an `ArgumentOutOfRangeException` so contract violations surface during development:

```csharp
if (payloadThresholdMB < MinPayloadThresholdBytes / (1024 * 1024))
    throw new ArgumentOutOfRangeException(nameof(payloadThresholdMB));
PayloadThresholdBytes = requestedBytes;
```

Activities/Python/UiPath.Python.Host.Shared/Program.cs
```diff
@@ -6,6 +6,7 @@ namespace UiPath.Python.Host
     internal static class Program
     {
         private static PythonService _service = null;
+        private static readonly CancellationTokenSource _cts = new CancellationTokenSource();
```
alexandru-petre, line 9: **[P3-Maintainability] MEDIUM — Host-side cancellation plumbing exists but isn't wired to any external signal**

The PR introduces this `_cts`, threads it into `RunServer` → `RunServerCore`, and uses `WaitAsync(ct)` to make the `ReadLineAsync` between-requests await cancellable. But the only place that ever calls `_cts.Cancel()` is `Application_ApplicationExit`, which runs as a `ProcessExit` handler. `ProcessExit` fires *after* `Main` returns — and `Main` can only return once `serverTask.GetAwaiter().GetResult()` unblocks, which requires `RunServerCore` to already have exited. So in practice the `_cts` is never the *cause* of `RunServerCore` exiting; it can only be a redundant signal once it's already exiting for another reason (client closed the pipe, IOException, unhandled exception).

It reads as "graceful host shutdown is now supported," but the new plumbing doesn't deliver that — there's no surface that an external party can use to ask the host to stop. The host's between-request wait remains effectively unbounded: if the client wedges (workflow hang, GC pause, parent process freeze), the host sits on `ReadLineAsync` indefinitely, holding a kernel pipe handle and a child process, until either the client disconnects (pipe IOException) or the host is killed externally.

**Suggested fix** — pick one of:
1. **Wire `_cts` to a real external signal.** Most natural here is parent-liveness: periodically check `Process.GetProcessById(parentPid).HasExited` (or open a handle to the parent and watch it) and call `_cts.Cancel()` when the parent is gone. Prevents orphan hosts when Studio/Robot crashes without closing the pipe. The parent PID can be passed as a startup arg from `Controller.CreateProcessStartInfo`.
2. **Or wire `_cts` to `Console.CancelKeyPress`** for parity with normal Ctrl+C semantics (lower value since the host is windowless in production, but cheap).
3. **Or remove the unwired plumbing** and accept that host shutdown is driven solely by the `Shutdown` request or the pipe IOException path.

Whichever path you pick, add a short comment explaining it — otherwise the next person to touch this code will read the `_cts`, the `WaitAsync(ct)`, and the `OperationCanceledException` handler and assume there's a wakeup signal that there isn't.

alexandru-petre (commented):
## Code Review — PR #555 (fix/python_issues)

**Verdict: NEEDS CHANGES**

| Priority | Count | Verdict Impact |
|----------|-------|----------------|
| P0-Correctness | 0 | — |
| P1-Bugs | 1 (LOW) | PASS |
| P2-Compatibility | 2 (1 MEDIUM, 1 LOW) | NEEDS CHANGES |
| P3-Maintainability | 5 (1 HIGH, 2 MEDIUM, 2 LOW) | NEEDS CHANGES |
| P4-Performance | 2 (1 MEDIUM, 1 LOW) | PASS WITH NOTES |
| P5-Security | 1 (MEDIUM) | PASS WITH NOTES |

Inline findings below.

### [P3-Maintainability] HIGH — No regression tests for new behavior

The PR adds substantial new logic — `ScriptDataSizeLimitMB` validation (design-time + runtime), `LogTraces` wiring through `EngineProvider` → `OutOfProcessEngine` → `HostWrapper`, payload-size enforcement in `PythonProxy.SendRequest`, `ResolveEffectiveVersion`, `ResultStateFor` mapping, generalized list/dictionary serialization in `Argument`, exception propagation through `RunServerCore`, host stdout/stderr buffering with rotation cap. The only test change is updating a delegate arity from 6 to 8 underscores. The PR description acknowledges this and defers to STUD=80106.

Every behavior in this PR is invisible to the existing test suite. A future regression (off-by-one in payload check, broken enum mapping, log-file leak, wrong dictionary roundtrip) will not be caught by CI. The "too risky for this release" reasoning argues for *more* coverage on the parts that don't require real Python, not less.

**Suggested cases** (none requires a Python install):
- `PythonScopeOptions.ScriptDataSizeLimitMB = 0` → `ArgumentOutOfRangeException`
- `Argument` round-trip for `List<int>`, `Dictionary<string,int>`, `byte[]`, `string`, `null`
- `PythonProxy.SendRequest` with payload exceeding threshold throws with both sizes formatted
- `ResultStateFor` returns correct `ResultState` for each `RequestType` (and `RequestException` for `null`)
- `HostWrapper.LogTrace` toggle: appending writes a file; repeated open failures don't crash; dispose closes the writer
- `EngineProvider.ResolveEffectiveVersion` with `Version=Auto` + empty path returns `Auto, autodetected=Auto`; with `Version=Python_310` + missing exe rethrows

### Positive observations

- Exception propagation rewrite in `PythonProxy.SendRequest`/`ReadResponse` (replacing `catch { /*ignore*/ }`) is a clear win.
- `RunServerCore` cleanly distinguishes fatal-exit exceptions (`OperationCanceledException`, `IOException`, `ObjectDisposedException`) from recoverable per-request errors; `ResultStateFor` mapping is the right shape.
- `RunServer()` `async void` → `async Task` is the right fix for the long-standing host-exit hazard; `Main`'s try/finally around `serverTask.GetAwaiter().GetResult()` with cts disposal is correctly ordered.
- `HostWrapper`'s post-mortem buffer (cap 512 lines, drained in `ThrowIfProcessHasExited`) is a real improvement over the prior `ReadToEnd()` approach, which could deadlock if the host filled its stdout buffer.
- `Argument.WrapForSerialization` / `Unwrap*` is well-factored — clear single-purpose helpers per shape, and `ChangeTypeOrNull` correctly handles null inside collections.
- Docs in `PythonScope.md` and `coded-api.md` updated in the same PR with explicit "leave unset in production" guidance — good discipline.

Activities/Python/UiPath.Python/EngineProvider.cs
alexandru-petre: I actually don't think this finding is important, the method is public but not part of the officially supported API and we can make the changes. I left Claude mentioning it just to agree on it.

Activities/Shared/UiPath.Shared.Service/Enums.cs
viogroza: the values were not used
they are used for internal purposes only

Activities/Python/UiPath.Python.Activities/PythonScope.cs
viogroza: will do, ty

Activities/Shared/UiPath.Shared.Service/Client/HostWrapper.cs
viogroza: will do, ty

Activities/Shared/UiPath.Shared.Service/Client/Controller.cs
viogroza: will do, ty

Activities/Python/UiPath.Python/Service/PythonProxy.cs
viogroza: this is intentional, leave it as it is
reason: payloadThresholdMB is for the user to allow passing of the params; a value to small, for sure is a mistake and we should ignore it; 
the real reason for this param is that too large payloads ca broke the pipe and the error is cryptic

Activities/Python/UiPath.Python.Host.Shared/Program.cs
viogroza: The expectation is that the server will receive a shutdown signal
I agree that it's worth considering sending the timeout param when the server process is starting, but at this moment I see it redundant (due to the shutdown signal)
the intention is to have the server up and running until it's specifically terminated (otherwise there can be cryptic/generic errors reported)

Activities/Python/UiPath.Python.Activities.API/Services/PythonService.cs
viogroza: will do, ty

Activities/Python/UiPath.Python/Properties/UiPath.Python.Designer.cs
viogroza: will revert, ty

viogroza:
The unittests will be addressed by another ticket (see in the description)

Activities/Python/UiPath.Python/Service/Argument.cs
```diff
@@ -16,55 +18,168 @@ public class Argument
...
-                return typedArray;
             }
 
-            //fix for Json serialize issue (ex: byte -> int)
-            if (!type.IsArray && type != _wrappedValue.GetType())
+            // fix for Json serialize issue (ex: byte -> int)
+            if (type != _wrappedValue.GetType())
             {
                 _wrappedValue = Convert.ChangeType(_wrappedValue, type);
             }
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Python/UiPath.Python/Service/Argument.cs
```diff
@@ -16,55 +18,168 @@ public class Argument
...
+            {
+                typedList.Add(ChangeTypeOrNull(item, elementType));
+            }
+            return typedList;
+        }
+
+        private static object ChangeTypeOrNull(object value, Type targetType)
+        {
+            return value == null ? null : Convert.ChangeType(value, targetType);
+        }
```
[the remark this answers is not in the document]
viogroza: will solve, ty

Activities/Python/UiPath.Python.Activities.API/Services/PythonService.cs
```diff
@@ -55,7 +60,30 @@ public async Task<IPythonScopeHandle> UsePythonScope(PythonScopeOptions options,
...
+            // Python 3.10 requires an explicit library file (python**.dll on Windows,
+            // libpython*.so on Linux).
+            if (effectiveVersion == Version.Python_310 &&
+                (string.IsNullOrWhiteSpace(options.LibraryPath) || !File.Exists(options.LibraryPath)))
+            {
+                throw new FileNotFoundException(string.Format(Resources.InvalidLibraryPathException, options.LibraryPath));
+            }
+
+            int payloadThresholdMB = options.ScriptDataSizeLimitMB ?? EngineProvider.DefaultPayloadThresholdMB;
+            IEngine engine = _engineFactory(options.Version, path, options.LibraryPath, false, options.Target, false, options.LogTraces, payloadThresholdMB);
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Shared/UiPath.Shared.Service/Client/HostWrapper.cs
```diff
@@ -72,16 +152,134 @@ internal bool GetHostProcessId(out int processId)
...
+                catch (Exception ex)
+                {
+                    Trace.TraceWarning($"Could not enforce Python host log retention: {ex.Message}");
                 }
             }
         }
+
+        // Test/integration hook: when set, this folder is used as the base.
+        // this in initialization and reset to null on teardown. The "python" subfolder is
+        // still appended on top.
```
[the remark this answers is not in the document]
viogroza: will do, ty

alexandru-petre (approved):

Activities/Python/UiPath.Python/Service/Argument.cs
```diff
@@ -16,55 +18,181 @@ public class Argument
...
+                {
+                    return UnwrapDictionary(array, keyType, valueType);
+                }
+                if (TryGetEnumerableElementType(type, out Type elementType))
                 {
-                    typedArray.SetValue(Convert.ChangeType(array.GetValue(i), type.GetElementType()), i);
+                    return UnwrapList(array, elementType);
                 }
-                return typedArray;
             }
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Python/UiPath.Python.Activities.API/Services/PythonService.cs
```diff
@@ -55,7 +60,30 @@ public async Task<IPythonScopeHandle> UsePythonScope(PythonScopeOptions options,
...
+
+            // Python 3.10 requires an explicit library file (python**.dll on Windows,
+            // libpython*.so on Linux).
+            if (effectiveVersion == Version.Python_310 &&
+                (string.IsNullOrWhiteSpace(options.LibraryPath) || !File.Exists(options.LibraryPath)))
+            {
+                throw new FileNotFoundException(string.Format(Resources.InvalidLibraryPathException, options.LibraryPath));
+            }
+
+            int payloadThresholdMB = options.ScriptDataSizeLimitMB ?? EngineProvider.DefaultPayloadThresholdMB;
```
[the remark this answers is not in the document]
viogroza: this is intentional, it will be coverted by another PR

Activities/Python/UiPath.Python/EngineProvider.cs
```diff
@@ -56,12 +59,41 @@ public static IEngine Get(Version version, string path, string libraryPath, bool
...
+            try
+            {
+                Autodetect(path, out autodetected);
+            }
+            catch
+            {
+                if (selected != Version.Auto) throw;
+            }
+            return selected != Version.Auto ? selected : autodetected;
+        }
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Python/UiPath.Python/Service/Argument.cs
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/Python/UiPath.Python.Activities.API/Services/PythonService.cs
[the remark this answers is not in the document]
viogroza: not an issue

Activities/Python/UiPath.Python/EngineProvider.cs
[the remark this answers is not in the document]
viogroza: will do, ty

viogroza:
improved serialization for a few types (instead of defaulting to List<T>)

PR #562, opened 2026-05-29
[STUD-79136]: Configurable PGP key size, IResource inputs, paired-property normalization

alexandru-petre:
## Problem Statement

The Cryptography pack accumulated several pain points since the v2.0 release:

1. **PGP key generation hardcoded to RSA-1024**, considered weak since 2010 (STUD-79136 calls for configurable key size, enterprise-grade default).
2. **No AEAD ciphers** — only legacy CBC modes; modern interop expects AESGCM / ChaCha20-Poly1305.
3. **Symmetric-cipher wire format incompatibility with third-party PGP tooling** — STUD-64429 surfaced cases where ServiceNow / devglan / OpenSSL ciphertext couldn't be decrypted (and vice versa) due to the deliberate PBKDF2-salt + random-IV format.
4. **No storage-resource (`IResource`) inputs** — file activities only accepted local string paths.
5. **Three different patterns** for path ↔ IResource and string ↔ SecureString toggles across the pack (switch-strict, switch-fallback, HasValue-inference); `[OverloadGroup]` on the existing PGP sign / clear-sign / verify activities was misused (validator rejected any XAML that bound both file + key inputs).
6. **NuGet packaging emits deprecation warnings** on every build (NU5125, NU5048).
7. **Activity class names mismatch display names** for the PGP activities, hurting Studio's search experience.
8. **PGP was silently treated as FIPS-compliant** in `IsFipsCompliant`, even though the underlying PgpCore + BouncyCastle path is not a CMVP-validated module. The dropdown label and design-time warning didn't surface this either.
9. **Coded-workflow service was incomplete and inconsistent** — `ICryptographyService` lacked the new `KeySize` parameter, had no way to validate a public key, exposed PGP keys as `Stream` (awkward for callers), had a partial Bytes/Text/File matrix (some forms missing per operation), and `CryptographyHelper` was exposed `public` as a quasi-API even though it's an internal implementation detail.
10. **Per-activity docs had drifted from source** — `PgpGenerateKeys.md` listed a non-existent `Password` property and missed `KeySize` entirely; multiple docs documented `[Browsable(false)]` properties; required-status markers didn't match `[RequiredArgument]` attributes.

## Analysis

The work splits into a few intertwined themes across the ~50 Cryptography-scoped commits on this branch. The PGP signing primitives themselves (Sign / Clearsign / Verify / GenerateKey) were already on develop via PR #514; this PR makes them production-ready, unifies the activity surface, ships a complete coded-workflow API, and synchronizes the docs.

### Activity-layer themes

- **PGP hardening (STUD-79136)** — `RsaKeySize` enum (2048 / 3072 / 4096) defaulting to 4096; SHA-256 hashing; AES-256 session keys; renamed `PgpGenerateKeyPair.Username → UserId` to match OpenPGP convention.
- **AEAD cipher modernization** — added AESGCM and ChaCha20-Poly1305; marked the older CBC ciphers obsolete; new default for `EncryptionAlgorithm` is AESGCM.
- **Symmetric interop diagnostics (STUD-64429)** — documented the wire format (`docs/symmetric-wire-format.md`) and emit a localized hint pointing at the spec when foreign-format ciphertext fails to decrypt.
- **IResource storage inputs** — paired path / `IResource` properties on every file activity (Encrypt / Decrypt / Hash / Pgp*) with viewmodel menu-action toggles and runtime resolution.
- **SecureString passphrase support** — paired plain / secure passphrase properties on all six PGP-passphrase-bearing activities.
- **`OverloadGroup` fix on PGP sign / clear-sign / verify** — distinct `[OverloadGroup]` names made the file + key pairs mutually exclusive at the validator level, which broke every valid XAML. Replaced with runtime fallback resolution.
- **Paired-property normalization** — every pair across the pack now follows one template: no dedicated `*InputModeSwitch` enum, viewmodel uses pure `HasValue`-inference for initial mode, runtime resolves via string-first fallback. Kept obsolete switches that existed on `origin/masters/Cryptography` for XAML deserialization compatibility; deleted switches that were only ever added since the last release.
- **PGP activity renames** for Studio search alignment: `PgpGenerateKeyPair → PgpGenerateKeys` (matches display "PGP Generate Keys"), `PgpClearSignFile → PgpClearsignFile` (matches GnuPG `--clearsign` spelling and RFC 9580 "cleartext signature"). `PgpSignFile` and `PgpVerify` already matched.
- **FIPS classification correctness** — `IsFipsCompliant(PGP)` now returns `false` (PgpCore + open-source BouncyCastle is not a CMVP-validated module). The dropdown label changed `PGP (Pretty Good Privacy)` → `PGP - Pretty Good Privacy (Non-FIPS)` to match how other non-FIPS ciphers are labeled, and the existing yellow design-time `FipsComplianceWarning` now also surfaces when PGP is selected (the `Algorithm != PGP &&` short-circuit was removed from the four CacheMetadata methods).
- **Activity-source `[RequiredArgument]` additions** — `PgpSignFile.OutputFilePath` and `PgpClearsignFile.OutputFilePath` (the only properties without a paired alternative). Other runtime-validated properties stay attribute-free because they have a `[Browsable(false)] IResource` paired alternative — adding `[RequiredArgument]` would force false design-time errors when the user switches to the resource alternative via the menu action.
- **Packaging cleanup** — removed deprecated `<PackageLicenseUrl>` / `<PackageIconUrl>` in favor of `<PackageLicenseFile>` (the existing `LICENSE.md` now ships in the .nupkg) and `<PackageProjectUrl>`.

### Coded-workflow API themes (`ICryptographyService`)

These methods were never released, so changes here aren't ABI-breaking:

- **Bytes / Text / File matrix** — every operation now exposes the three input/output forms (`*Bytes` / `*Text` / `*File`) with consistent suffixes; 23 new methods filling previously-missing cells (symmetric `EncryptBytes` / `DecryptBytes`, PGP `*File` variants for Encrypt/Decrypt, PGP `*Text` variants for Sign/Clearsign, PGP `*Text` / `*File` variants for Verify/VerifyClear/VerifyPublicKey, plus `KeyedHashBytes`).
- **`KeySize` propagation** — `PgpGenerateKeys` now exposes the 5-arg overload with `RsaKeySize keySize = RsaKeySize.Rsa4096` default; the previous 4-arg overload collapses into it via default-value resolution.
- **`PgpVerifyPublicKey*` family** — previously unreachable from coded workflows; now available in Bytes / Text / File forms.
- **Stream → byte[]** — PGP public/private key parameters are now `byte[]` rather than `Stream`, removing the disposal-ownership ambiguity and matching what coded callers naturally have (from `File.ReadAllBytes`, HTTP responses, Storage Buckets). The byte form accepts both ASCII-armored and binary OpenPGP encodings.
- **Method-name normalization** — byte-form methods carry an explicit `Bytes` suffix (`EncryptBytes`, `PgpSignBytes`, etc.), mirroring the existing `Text` and `File` suffixes; overloads of the same logical method are grouped adjacent in both the interface and the implementation.
- **Default-value consistency** — file methods carry the same `overwrite = false` / `sign = false` / `verifySignature = false` defaults as their bytes counterparts, eliminating overloads that existed only to provide a missing default.
- **`CryptographyHelper` soft-hide** — class is now `[Obsolete]` + `[EditorBrowsable(EditorBrowsableState.Never)]`, pointing external callers at `ICryptographyService`. Couldn't be fully internalised because `UiPath.Shared` is a shared project compiled into multiple assemblies, so `InternalsVisibleTo` would cause duplicate-type ambiguity on every shared utility (documented in the class XML doc).
- **New `CryptographyHelper` helpers** — `PgpSignText`, `PgpClearSignText`, `PgpVerifyText`, `PgpVerifyClearText` (armored-string variants via PgpCore's `SignArmoredString` / `VerifyArmoredString` etc.) so the service's `*Text` PGP methods don't have to fake armoring on top of byte methods.

### Documentation themes

- **Per-activity markdown regenerated** from source via the `activity-doc-generator` skill — fixed the drift surfaced by an adversarial audit: `PgpGenerateKeys.md` now correctly documents `Passphrase` / `PassphraseSecureString` (instead of the non-existent `Password`), `KeySize` is documented (default `Rsa4096`), `[Browsable(false)]` properties are no longer leaked, paired-property primaries are marked Conditional with a note about the IResource alternative, `Overwrite` columns now surface the `[RequiredArgument]` they carry, and `PgpVerify.md` no longer claims `Mode` is required.
- **Skill validator bug fix** — the `activity-doc-generator` validator's single-backtick `xml` substring check fired inside legitimate triple-backtick fences and made it impossible to pass any doc; tightened to a negative-lookbehind regex.
- **Test cleanup robustness** — moved temp-file creation inside `try` blocks across 9 PGP/symmetric tests so the `finally` cleanup is reachable even if the create-step itself throws.

## Considered Use Cases

- **PGP key generation**: RSA-2048 / 3072 / 4096; legacy XAML omitting `KeySize` defaults to 4096.
- **Symmetric encryption**: AESGCM / ChaCha20-Poly1305 (new), AES / DES / RC2 / Rijndael / TripleDES (obsolete but still functional for back-compat).
- **PGP encrypt / decrypt** with optional sign / verify; passphrase as string or SecureString.
- **PGP sign / clearsign / verify**: binary signature, clearsign cleartext signature, public-key validation modes.
- **IResource inputs**: storage-bucket resource provided in place of a local file path (input file, public / private key file).
- **Three customer workflows reported during development**: full sign → clearsign → verify pipeline; EncryptText + DecryptText with sign + verify; EncryptFile + DecryptFile with sign + verify — all three covered by new end-to-end regression tests.
- **Interop with third-party PGP tooling** (STUD-64429): runtime error now hints at the wire-format spec instead of producing an opaque `CryptographicException`.
- **FIPS-locked machines**: PGP now visibly flagged as non-FIPS in the dropdown and triggers the standard design-time warning, so operators can make an informed choice before deployment.
- **Coded workflows**: full Bytes/Text/File matrix per operation, both passphrase forms, every PGP verify mode reachable, key generation with custom key size — all covered by 78 service-level tests using a shared key-pair fixture.

## Implementation

**New files**:
- `UiPath.Cryptography/Enums/RsaKeySize.cs` — 2048 / 3072 / 4096.
- `UiPath.Cryptography.Activities/Helpers/PgpFileResolver.cs` — shared string-first fallback for path ↔ IResource and string ↔ SecureString.
- `Activities/Cryptography/docs/symmetric-wire-format.md` — STUD-64429 spec.
- `UiPath.Cryptography.Activities.Tests/Resources/ActivityMetadataTests.cs` — metadata invariant checks.
- `PgpGenerateKeys.cs` (renamed + rewritten from `PgpGenerateKeyPair.cs`), `PgpClearsignFile.cs` (renamed + rewritten from `PgpClearSignFile.cs`), and the matching viewmodels.

**Coded API** (`UiPath.Cryptography.Activities.API`):
- `ICryptographyService.cs` — reorganized; PGP key params now `byte[]`; full Bytes/Text/File matrix; `Bytes` suffix on byte-form methods; defaults harmonized.
- `Services/CryptographyService.cs` — matching implementation with `MemoryStream` adapters into the helper layer.
- `UiPath.Cryptography.Activities.API.Tests/CryptographyServiceTests.cs` — 78 tests covering validation + roundtrip for every public method, with a shared `PgpKeyFixture` (RSA-2048 for test speed).

**Activity docs**:
- `UiPath.Cryptography.Activities.Packaging/docs/activities/*.md` — 10 per-activity docs regenerated from source.

**Modified** (~60 files):
- All file-based Encrypt / Decrypt / Hash / PGP activities — IResource pair properties, OutputFilePath fallback to input directory, sentence-case display-name fixes, paired-property template normalization, uniform `IsFipsCompliant` check (no PGP short-circuit), `[RequiredArgument]` on the two `OutputFilePath`s without paired alternatives.
- ViewModels — pure `HasValue`-inference for initial mode, switch handlers null the inactive side, no writes to obsolete `*InputModeSwitch` enums.
- `CryptographyHelper.cs` — added AESGCM / ChaCha20-Poly1305; `PgpSignText` / `PgpClearSignText` / `PgpVerifyText` / `PgpVerifyClearText` helpers; documented obsolete-cipher rationale; corrected `IsFipsCompliant(PGP)` to return `false`; class marked `[Obsolete]` + `[EditorBrowsable(Never)]` with explanatory XML doc.
- In-package consumers of `CryptographyHelper` — `#pragma warning disable CS0618` at file scope (matches the existing pattern used for obsolete encryption-algorithm enum values).
- `UiPath.Cryptography.resx` — `PGP` label gets the `(Non-FIPS)` suffix, parens around the long form replaced with a hyphen for legibility.
- `ActivitiesMetadata.json`, English `.resx`, `Designer.cs` accessors — kept in lock-step with class / property renames and switch deletions.

**Removed**: deprecated `<PackageLicenseUrl>` / `<PackageIconUrl>` in the Packaging csproj.

## Caveats / Potential Issues

- **`IsFipsCompliant(PGP)` semantics flipped** from `true` to `false`. This now produces an observable yellow design-time warning when the algorithm is selected — that's the intended behavior, but it is a new warning that didn't appear in v2.0 for workflows that had been authored against PGP.
- **`CryptographyHelper` is `[Obsolete]`** with warning severity (not error). Anyone who took a direct dependency on this class — which was technically possible since it's `public` and bundled in the activities .nupkg — now sees a CS0618 warning at compile time pointing them at `ICryptographyService`. The class continues to function; no ABI break. We can't make it fully internal because `UiPath.Shared` is a shared project (`.shproj`) compiled into multiple assemblies; `InternalsVisibleTo` would cause duplicate-type ambiguity on `System.Extensions.ToComparer<T>` and similar shared utilities.
- **Locale `.resx` files are intentionally stale** for new keys added on this branch (Passphrase / Passphrase Secure, IResource displays, UserId, normalized pair entries). The resource manager falls back to English for missing keys. A translation pass is a follow-up.
- **Kept-obsolete switches** (`KeyInputModeSwitch` on Encrypt / Decrypt / Hash, `FileInputModeSwitch` on Encrypt / Decrypt) are still declared on the activity class for XAML back-compat with already-published workflows. They're `[Browsable(false)] [Obsolete]` and never read by the viewmodel or runtime — deserialization is by construction (property exists on the class), not verified by a load-an-XAML test.
- **Async PGP sign / clearsign / verify**: no design-time validation for the file path ↔ IResource pair (runtime fallback handles it). Matches the rest of the pack now.
- **`OutputFileName` + `OutputFilePath` both set**: pre-existing behavior preserved (FileName silently ignored when FilePath is also set). Documented as an open item; no reported customer impact.

## How to Test

1. `dotnet build Activities/Activities.Cryptography.sln` — expect 0 errors. NU5125 / NU5048 should be gone.
2. `dotnet test Activities/Activities.Cryptography.sln` — expect 595 / 595 passing (517 activity + 78 coded-API).
3. Targeted customer-workflow regressions: `dotnet test --filter "FullyQualifiedName~RoundTrip|FullyQualifiedName~EndToEnd"` — covers the three customer XAMLs reported during development.
4. `python .claude/skills/activity-doc-generator/scripts/validate-activity-docs.py Activities/Cryptography/UiPath.Cryptography.Activities.Packaging/docs --strict` — 12 markdown files validated, 0 findings.
5. Manual smoke in Studio:
   - Drop **PGP Generate Keys**; verify the KeySize dropdown shows 2048 / 3072 / 4096 and defaults to 4096.
   - Sign → Verify with **PGP Sign File** + **PGP Verify** (binary mode).
   - Clearsign → Verify with **PGP Clearsign File** + **PGP Verify** (Clearsigned File mode).
   - Search the activity panel for `PGP` and `Generate Keys` — both should surface the activity (renames make the class names match the display names).
   - On **Encrypt File** (or any of the four symmetric activities), open the Algorithm dropdown — `PGP - Pretty Good Privacy (Non-FIPS)` should be visible. Select it: the yellow `FipsComplianceWarning` should appear in the designer next to the activity.
6. Manual smoke for coded workflows: in a Coded C# project, instantiate `ICryptographyService` and confirm IntelliSense shows the full Bytes/Text/File matrix; confirm `CryptographyHelper` does not appear in IntelliSense.


viogroza:
there are some conflicts, please rebase

viogroza:
claude has some findings (on aggregate for all the 74 commits present at the time)
at least some of them seem accurate
[code-review-cryptography-pgp-keysize-resource.md](https://github.com/user-attachments/files/28594044/code-review-cryptography-pgp-keysize-resource.md)
I will do additional checks and post additional comments

viogroza:
also, upgrading an existing workflow causes it to break (unresolved activity); 
I created project with cryptography activities 2.0.0 and upgraded to a custom version built from this branch
(this is a xaml file, but I renamed the extension to .txt - it seems that uploading directly .xaml files is not permitted)
[TestsCrypto2.0 - Copy.txt](https://github.com/user-attachments/files/28594268/TestsCrypto2.0.-.Copy.txt)

HashFile doesn't load properly
image

Activities/Credentials/UiPath.Credentials.Activities.ViewModels/Resources/ActivitiesMetadataWindows.json
```diff
@@ -1,309 +1,309 @@
...
+        {
+          "name": "Result",
+          "displayNameKey": "Result",
+          "isRequired": false,
+          "isVisible": true,
+          "isPrincipal": false,
+          "category": {
+            "name": "Output",
+            "displayNameKey": "Output"
+          }
```
viogroza: it seems that only the indentation has changed for this file

Activities/Cryptography/UiPath.Cryptography.Activities.Tests/UiPath.Cryptography.Activities.Tests.csproj
```diff
@@ -17,9 +17,11 @@
 		<PackageReference Include="UiPath.Activities.Contracts" PrivateAssets="All" />
 		<PackageReference Include="UiPath.Activities.Api" PrivateAssets="All" />
 		<PackageReference Include="PgpCore" />
+		<PackageReference Include="Newtonsoft.Json" Version="13.0.3" />
```
viogroza: the version should not be hardcoded here (it should be inherited from the .props or .targets

viogroza:
I did a sanity check on the activities and it looks good so far.. 
I was able to add the activities and run them

Activities/Cryptography/UiPath.Cryptography/CryptographyHelper.cs
```diff
@@ -184,22 +227,23 @@
...
+                    return new AesCryptoServiceProvider(); // kept for backwards compatibility
 
                 case EncryptionAlgorithm.AESGCM:
-                    throw new InvalidOperationException(Resources.UnsupportedSymmetricAlgorithmException); //it's implemented separately.
+                case EncryptionAlgorithm.ChaCha20Poly1305:
+                    throw new InvalidOperationException(Resources.UnsupportedSymmetricAlgorithmException); //implemented separately as AEAD.
 
                 case EncryptionAlgorithm.DES:
-                    return new DESCryptoServiceProvider();
+                    return new DESCryptoServiceProvider(); // kept for backwards compatibility
```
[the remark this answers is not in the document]
alexandru-petre: Must be kept for backwards compatibility, we are warning users if they are still using it.

viogroza:
new review from claude (most of the previous findings are addressed)
**the important one - owner decision - please review**

## Remaining action list (post-follow-up)

**Owner decision (blocking for release semantics):**
1. **Rename back-compat policy** (Medium) — confirm the breaking renames + major-version bump / migration note with repo owners (CLAUDE.md requirement). Only open item of real weight.

**Nice-to-have before merge (all low effort):**
2. Add `SendWithException(ex)` + `Trace.TraceError` to `PgpVerify` error path (and bind `ex`) — consistency with siblings.
3. Add a test asserting the SHA-256 signing default (closes the only High-fix coverage gap).
4. Remove or exercise dead `TestLocalResource`.
5. Rename `PgpGenerateKeys_WithKeySize_CreatesValidKey` (or assert modulus size).

**Backlog (optional):**
6. Migrate `PackageIconUrl` → embedded `<PackageIcon>` (NU5048 deprecation; pre-existing).
7. Add a persisted-PGP-mode open test for ViewModel field visibility.
8. Consider hoisting the `ChaCha20Poly1305.IsSupported` check to `CacheMetadata` for earlier/design-time feedback.

viogroza:
@alexandru-petre 
Update problem is solved
Claude says ok:
 Bottom line: this commit clears essentially every remaining actionable finding

  Fixed in ea47f8c (11 files, +193/−56):
  - ✅ PgpVerify error-path telemetry — catch (Exception ex) now calls SendWithException(ex) + Trace.TraceError, matching the established KeyedHashFile pattern. Telemetry op hoisted out of try; the using is now
  unconditional. Confirmed compile-safe — ITelemetryOperationWrapper lives in the always-compiled UiPath.Shared.Activities shared project.
  - ✅ SHA-256 signing test (the round-1 High-fix coverage gap) — new PgpSign_DefaultsToSha256 parses the signature packet and asserts HashAlgorithmTag.Sha256.
  - ✅ ChaCha20Poly1305 IsSupported in CacheMetadata — added as a design-time warning to all four symmetric activities (so it's now guarded at both design time and runtime).
  - ✅ CreatesValidKey test — now actually asserts masterKey.BitStrength == (int)keySize, so the name is justified.
  - ✅ Dead TestLocalResource — removed, with its orphaned usings.
  - ✅ Persisted-open visibility — new PairedInputToggleTests.cs (5 facts) covers the UseSecondary contract that drives field visibility on reopen, including the IResource pair; InternalsVisibleTo added
  cleanly.

  Still open (not code defects):
  - ⚠️ Rename back-compat policy (Medium) — unchanged; needs repo-owner sign-off + major-version/migration note per CLAUDE.md. This is the sole gating item before merge.
  - ⚠️ PackageIcon embedded migration — backlog, acceptable.

  One trivial new nit: PairedInputToggle still carries [ExcludeFromCodeCoverage] even though it now has a dedicated test suite — consider dropping the attribute so the coverage counts.

  No new correctness defects found in the fresh pass. The branch is in good shape to merge once the rename policy is confirmed with owners.

viogroza (approved):

PR #525, opened 2026-03-20
cryptography: Add AI-generated XAML activity docs for Cryptography package [STUD-79287]

AlexMarinescuUiPath:
## What changed?
Added structured per-activity markdown reference files for all 10 activities in the UiPath Cryptography package, plus a package overview doc. Coverage includes:

- \EncryptText\, \DecryptText\, \EncryptFile\, \DecryptFile\
- \KeyedHashText\, \KeyedHashFile\
- \PgpClearSignFile\, \PgpSignFile\, \PgpVerify\, \PgpGenerateKeyPair\

Each file documents inputs, outputs, configuration options, and includes a copy-paste XAML example. Files are co-located under \UiPath.Cryptography.Activities.Packaging/docs/\.

## Why?
Enables LLM CLI tooling (Claude Code and similar agents) to use these docs as grounded activity references when generating or reviewing automation workflows, reducing hallucinations and improving accuracy for Cryptography activities.

## Jira Link
[STUD-79287]

## Dependencies
- [x] None

[STUD-79287]: https://uipath.atlassian.net/browse/STUD-79287?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

PR #542, opened 2026-04-28
feat: add coded workflow API support for all activity packages [STUD-79807]

viogroza:
Introduces UiPath.{Category}.Activities.API projects for Cryptography, Database, FTP, Java, Python, and Credentials. 
Each package exposes a typed service interface (ICryptographyService, IDatabaseService, IFtpService, IJavaService, IPythonService, ICredentialsService) and registers it with the coded workflows service locator via ICodedWorkflowsServiceRegistry, making the services available as injected dependencies in coded workflow classes.

Scope-based packages (FTP, Java, Python) additionally expose IXxxScopeHandle types and static extension operations classes (FtpOperations, JavaOperations, PythonOperations) for fluent use within a session.

All API projects include unit test coverage and are wired into every per-category solution file and the community solution.

**Fixed nuget config**

Added/Updated dependencies
UiPath.Platform 25.12.1
UiPath.Telemetry.Client 2.9.1
UiPath.CodedWorkflows 24.10.2

https://uipath.atlassian.net/browse/STUD-79807

TO DO

- [x] Add coded .md files
- [ ] Add coded samples in this repo https://github.com/UiPath/codedautomations-samples/tree/main

PR #569, opened 2026-06-09
feat(cryptography): third-party symmetric interoperability [STUD-64429][STUD-80231][STUD-80425]

alexandru-petre:
## Context
The Cryptography activity package (`Activities/Cryptography/`) — symmetric encrypt/decrypt activities (EncryptText, DecryptText, EncryptFile, DecryptFile) and the coded-workflow `CryptographyService`. Consumed from XAML workflows via the Studio toolbox and from coded workflows via `ICryptographyService`. Internally delegates to `CryptographyHelper` for cipher operations.

## Problem Statement
UiPath's symmetric ciphertext layout (`salt(8) || IV || ct [|| tag(16)]`, PBKDF2-HMAC-SHA1 @ 10,000 iter, max-legal key size) is UiPath-specific. Output from `EncryptText`/`EncryptFile` cannot be read by openssl, Java javax.crypto, Python cryptography, ServiceNow, or browser SubtleCrypto; conversely, `DecryptText`/`DecryptFile` cannot consume blobs from those tools. The root customer report (STUD-64429) was TripleDES round-trip failure, but the same issue affects every algorithm and blocks every integration scenario.

## Behavior Before This PR
A workflow encrypts a payload with `EncryptText(Algorithm = AES, Key = "shared")` and hands the Base64 to an external system. `openssl enc -aes-256-cbc -pbkdf2 -k shared -d` rejects it (no magic prefix). The reverse direction also fails: openssl-produced blobs return `Decryption failed. UiPath wire format is Base64(salt | IV | ciphertext)…`. No interop path exists.

## Behavior After This PR
Same scenario, opting into `Format = OpenSslEnc`: output is `Salted__ || salt(8) || ct` with PBKDF2-HMAC-SHA256 @ 600,000 iter — what `openssl enc -aes-256-cbc -pbkdf2 -iter 600000 -d` expects. Classic stays the default and remains byte-stable (existing customer ciphertext still works; pinned by hex-literal fixtures). Three new opt-in formats: `Owasp2026` (Classic layout, caller-controlled iter), `Raw` (caller-supplied key + IV, no KDF — interops with .NET native AES, openssl `-K -iv`, KMS keys), `OpenSslEnc`.

The `CryptoKey` hierarchy is refactored so `(key kind × wire format)` pairings are enforced at compile time (`Classic(PasswordKey)`, `Raw(RawKey)`). `PasswordKey` bytes are now zeroed per call (P2 fix). KDF-iteration validation rejects negatives (P3 fix).

## Implementation
- **Classic is intentionally frozen.** It is the byte-stable contract for existing UiPath ciphertext. Future OWASP iter upgrades go into new enum entries (`Owasp2026` is a year-snapshot; a future revision would add `Owasp2030` rather than mutate the constant) — enforced by `WireFormatStabilityTests` decrypting pre-captured Classic blobs.
- **`ReleaseMaterialisedBytes` virtual on `CryptoKey`.** Defaults to no-op because `RawKey.KeyBytes` returns instance-owned storage; clearing it would corrupt the key. `PasswordKey` overrides to zero. Every `KeyBytes` access in the service and the four activities is wrapped in try/finally that calls it.

## Documentation
- `KeyedHash{File,Text}.md` Result description now correctly says "upper-case hexadecimal string" — `KeyedHash{File,Text}` always returned upper-case via `BitConverter.ToString` default, but the per-activity docs (added on `develop` by #525) claimed lower-case. Aligned with `coded-api.md` and the `ICryptographyService` XML doc-comments, which already said upper-case. Source code unchanged (STUD-80425).

## Caveats
- `OpenSslEnc` is hardcoded to AES-256 (max legal key size). `aes-128-cbc` / `aes-192-cbc` interop via the activity surface is tracked as **STUD=80390** (Improvement, follow-up). OpenSSL CLI tests only exercise `aes-256-cbc`; a code comment in `ExternalInteropOpenSslCliTests.cs` documents the limitation.
- AEAD over `OpenSslEnc` (AES-GCM + `Salted__` prefix) is a UiPath extension, not a cross-tool standard — `openssl enc` itself doesn't support GCM. Called out in `docs/symmetric-wire-format.md`.

## How to Test
1. `dotnet test Activities/Activities.Cryptography.sln` → 858 passing / 0 failing.
2. Interop-only subset: `dotnet test … --filter "FullyQualifiedName~Interop"`.
3. Wire-format stability: `dotnet test … --filter "FullyQualifiedName~WireFormatStability"` decrypts hex-literal blobs (any future layout change fails these).
4. Live external-tool interop (gated): `ExternalInteropOpenSslCliTests` / `ExternalInteropGpgCliTests` no-op when the CLI is missing; on agents with OpenSSL/GnuPG they round-trip against the reference implementation.
5. Manual: in Studio, create `EncryptText` with `Format = OpenSslEnc, Algorithm = AES`, encrypt; pipe Base64 → `openssl enc -d -aes-256-cbc -pbkdf2 -iter 600000 -k <password> -A -base64` and verify plaintext.

### DIFF
--- a/Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
@@ -1,43 +1,42 @@
 namespace UiPath.Cryptography.Activities.API
 {
     /// <summary>
-    /// Optional knobs for symmetric encrypt operations. Default-constructed value
-    /// produces <see cref="SymmetricWireFormat.Classic"/> output with auto-generated IV
-    /// and the format's recommended KDF iterations.
+    /// Options for symmetric encrypt operations. Bundles the key, wire format, and any
+    /// format-specific knobs (IV for <see cref="SymmetricWireFormat.Raw"/>, KDF iterations
+    /// for <see cref="SymmetricWireFormat.Owasp2026"/> / <see cref="SymmetricWireFormat.OpenSslEnc"/>).
+    /// Construct via a format factory; the factory's key parameter type enforces the
+    /// (key kind × wire format) pairing at compile time.
     /// </summary>
-    public sealed class SymmetricEncryptOptions
+    public sealed class SymmetricEncryptOptions : CryptoOptions
     {
-        /// <summary>
-        /// Wire format to produce. Defaults to <see cref="SymmetricWireFormat.Classic"/>
-        /// — the byte-stable UiPath layout PBKDF2-HMAC-SHA1 @ 10 000 iterations.
-        /// </summary>
-        public SymmetricWireFormat Format { get; private init; } = SymmetricWireFormat.Classic;
+        private SymmetricEncryptOptions() { }
 
         /// <summary>
         /// Explicit initialization vector for <see cref="SymmetricWireFormat.Raw"/>.
-        /// Null (default) means generate a random IV. Must be null for all non-Raw formats.
+        /// Null (default) means generate a random IV. Set only by the
+        /// <see cref="Raw(RawKey, byte[])"/> factory.
         /// </summary>
-        public byte[] Iv { get; private init; }
+        public byte[] IV { get; private init; }
 
-        /// <summary>
-        /// PBKDF2 iteration count override for <see cref="SymmetricWireFormat.Owasp2026"/>
-        /// and <see cref="SymmetricWireFormat.OpenSslEnc"/>. Zero (default) uses the
-        /// format's OWASP-recommended value. Must be zero for Classic and Raw.
-        /// </summary>
-        public int KdfIterations { get; private init; }
+        /// <summary>Produces <see cref="SymmetricWireFormat.Classic"/> output — UiPath's frozen, byte-stable layout (PBKDF2-HMAC-SHA1 @ 10 000 iter).</summary>
+        public static SymmetricEncryptOptions Classic(PasswordKey key) =>
+            new() { Key = key, Format = SymmetricWireFormat.Classic };
 
-        /// <summary>Produces <see cref="SymmetricWireFormat.Classic"/> output (UiPath's frozen, byte-stable layout).</summary>
-        public static SymmetricEncryptOptions Classic() =>
-            new() { Format = SymmetricWireFormat.Classic };
-
-        /// <summary>Produces <see cref="SymmetricWireFormat.Owasp2026"/> output (Classic layout, OWASP-recommended PBKDF2-HMAC-SHA1 iterations).</summary>
-        /// <param name="kdfIterations">Optional override. Zero uses 1,300,000 (OWASP 2026 recommendation).</param>
-        public static SymmetricEncryptOptions Owasp2026(int kdfIterations = 0) =>
-            new() { Format = SymmetricWireFormat.Owasp2026, KdfIterations = kdfIterations };
+        /// <summary>Produces <see cref="SymmetricWireFormat.Owasp2026"/> output (Classic wire layout, PBKDF2-HMAC-SHA1).</summary>
+        /// <param name="key">Password material; PBKDF2 stretches it to the cipher key.</param>
+        /// <param name="kdfIterations">
+        /// PBKDF2 iteration count. Defaults to <c>1_300_000</c> — the OWASP 2026 recommendation
+        /// for PBKDF2-HMAC-SHA1. The literal is part of the year-snapshot contract: if OWASP
+        /// revises the recommendation, this package adds a new <see cref="SymmetricWireFormat"/>
+        /// entry (e.g. <c>Owasp2030</c>) rather than changing this default.
+        /// </param>
+        public static SymmetricEncryptOptions Owasp2026(PasswordKey key, int kdfIterations = 1_300_000) =>
+            new() { Key = key, Format = SymmetricWireFormat.Owasp2026, KdfIterations = kdfIterations };
 
         /// <summary>
         /// Produces <see cref="SymmetricWireFormat.Raw"/> output (caller-supplied key + IV, no KDF).
         /// </summary>
+        /// <param name="key">Literal cipher key of the algorithm's required size (e.g. 32 bytes for AES-256).</param>
         /// <param name="iv">
         /// Optional explicit IV. Null (the default) lets the cipher generate one.
         /// <para>
@@ -49,12 +48,19 @@ namespace UiPath.Cryptography.Activities.API
         /// third-party protocol mandates it, and ensure your producer guarantees uniqueness.
         /// </para>
         /// </param>
-        public static SymmetricEncryptOptions Raw(byte[] iv = null) =>
-            new() { Format = SymmetricWireFormat.Raw, Iv = iv };
+        public static SymmetricEncryptOptions Raw(RawKey key, byte[] iv = null) =>
+            new() { Key = key, Format = SymmetricWireFormat.Raw, IV = iv };
 
         /// <summary>Produces <see cref="SymmetricWireFormat.OpenSslEnc"/> output (<c>openssl enc</c>-compatible, PBKDF2-HMAC-SHA256).</summary>
-        /// <param name="kdfIterations">Optional override. Zero uses 600,000 (OWASP 2026 recommendation for SHA-256).</param>
-        public static SymmetricEncryptOptions OpenSslEnc(int kdfIterations = 0) =>
-            new() { Format = SymmetricWireFormat.OpenSslEnc, KdfIterations = kdfIterations };
+        /// <param name="key">Password material; PBKDF2-SHA256 stretches it to key+IV.</param>
+        /// <param name="kdfIterations">
+        /// PBKDF2 iteration count. Defaults to <c>600_000</c> — the OWASP 2026 recommendation
+        /// for PBKDF2-HMAC-SHA256. (Note: <c>openssl enc</c>'s own default is 10 000 for
+        /// back-compat; the OWASP-aligned default produces stronger output but is still
+        /// decryptable by <c>openssl enc -pbkdf2 -iter 600000 -md sha256</c>.) Pass <c>10_000</c>
+        /// to match the openssl back-compat default explicitly.
+        /// </param>
+        public static SymmetricEncryptOptions OpenSslEnc(PasswordKey key, int kdfIterations = 600_000) =>
+            new() { Key = key, Format = SymmetricWireFormat.OpenSslEnc, KdfIterations = kdfIterations };
     }
 }
--- a/Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
@@ -1,0 +1,60 @@
+namespace UiPath.Cryptography.Activities.API
+{
+    /// <summary>
+    /// Optional knobs for symmetric encrypt operations. Default-constructed value
+    /// produces <see cref="SymmetricWireFormat.Classic"/> output with auto-generated IV
+    /// and the format's recommended KDF iterations.
+    /// </summary>
+    public sealed class SymmetricEncryptOptions
+    {
+        /// <summary>
+        /// Wire format to produce. Defaults to <see cref="SymmetricWireFormat.Classic"/>
+        /// — the byte-stable UiPath layout PBKDF2-HMAC-SHA1 @ 10 000 iterations.
+        /// </summary>
+        public SymmetricWireFormat Format { get; private init; } = SymmetricWireFormat.Classic;
+
+        /// <summary>
+        /// Explicit initialization vector for <see cref="SymmetricWireFormat.Raw"/>.
+        /// Null (default) means generate a random IV. Must be null for all non-Raw formats.
+        /// </summary>
+        public byte[] Iv { get; private init; }
+
+        /// <summary>
+        /// PBKDF2 iteration count override for <see cref="SymmetricWireFormat.Owasp2026"/>
+        /// and <see cref="SymmetricWireFormat.OpenSslEnc"/>. Zero (default) uses the
+        /// format's OWASP-recommended value. Must be zero for Classic and Raw.
+        /// </summary>
+        public int KdfIterations { get; private init; }
+
+        /// <summary>Produces <see cref="SymmetricWireFormat.Classic"/> output (UiPath's frozen, byte-stable layout).</summary>
+        public static SymmetricEncryptOptions Classic() =>
+            new() { Format = SymmetricWireFormat.Classic };
+
+        /// <summary>Produces <see cref="SymmetricWireFormat.Owasp2026"/> output (Classic layout, OWASP-recommended PBKDF2-HMAC-SHA1 iterations).</summary>
+        /// <param name="kdfIterations">Optional override. Zero uses 1,300,000 (OWASP 2026 recommendation).</param>
+        public static SymmetricEncryptOptions Owasp2026(int kdfIterations = 0) =>
+            new() { Format = SymmetricWireFormat.Owasp2026, KdfIterations = kdfIterations };
+
+        /// <summary>
+        /// Produces <see cref="SymmetricWireFormat.Raw"/> output (caller-supplied key + IV, no KDF).
+        /// </summary>
+        /// <param name="iv">
+        /// Optional explicit IV. Null (the default) lets the cipher generate one.
+        /// <para>
+        /// <b>⚠ NEVER reuse the same (Key, IV) pair across encryptions.</b> Doing so destroys
+        /// confidentiality (CTR keystream reuse) and — under AEAD modes (AES-GCM,
+        /// ChaCha20-Poly1305) — additionally lets an attacker recover the authentication key
+        /// and forge arbitrary messages under that key. Prefer leaving <paramref name="iv"/>
+        /// null so a fresh random IV is generated per call; only supply an explicit IV when a
+        /// third-party protocol mandates it, and ensure your producer guarantees uniqueness.
+        /// </para>
+        /// </param>
+        public static SymmetricEncryptOptions Raw(byte[] iv = null) =>
+            new() { Format = SymmetricWireFormat.Raw, Iv = iv };
+
+        /// <summary>Produces <see cref="SymmetricWireFormat.OpenSslEnc"/> output (<c>openssl enc</c>-compatible, PBKDF2-HMAC-SHA256).</summary>
+        /// <param name="kdfIterations">Optional override. Zero uses 600,000 (OWASP 2026 recommendation for SHA-256).</param>
+        public static SymmetricEncryptOptions OpenSslEnc(int kdfIterations = 0) =>
+            new() { Format = SymmetricWireFormat.OpenSslEnc, KdfIterations = kdfIterations };
+    }
+}
--- a/Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
@@ -1,3 +1,4 @@
+using System;
 using System.Text;
 
 namespace UiPath.Cryptography.Activities.API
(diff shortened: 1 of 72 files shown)


Activities/Cryptography/UiPath.Cryptography.Activities.API/Services/ICryptographyService.cs
```diff
@@ -15,295 +13,276 @@ namespace UiPath.Cryptography.Activities.API
...
-        string DecryptText(string input, EncryptionAlgorithm algorithm, SecureString key, Encoding encoding);
-
-        /// <summary>Decrypts a Base64-encoded string using the specified algorithm and raw key bytes.</summary>
-        string DecryptText(string input, EncryptionAlgorithm algorithm, byte[] keyBytes, Encoding encoding);
+        /// <summary>
+        /// Encrypts arbitrary bytes using the algorithm and wire format selected on
+        /// <paramref name="options"/>. Returns the ciphertext — exact byte layout depends on
+        /// the format (see <see cref="SymmetricWireFormat"/>).
+        /// </summary>
+        byte[] EncryptBytes(byte[] input, EncryptionAlgorithm algorithm, SymmetricEncryptOptions options);
```
LiviuPonova, line 66: **Breaking change to the coded-workflow API with no migration path.** This rewrite removes all of the previous public overloads (~60: `string`/`SecureString`/`byte[]` key + `Encoding` variants, the path-based `PgpGenerateKeys`, …) with no `[Obsolete]` shims, so any coded workflow written against the API introduced in #542 fails to compile on package upgrade.

Given that API is only ~6 weeks old, the break itself may well be the right call — but two asks:
1. Explicit owner sign-off, per the repo policy on public-contract breaks.
2. A short migration section in `coded-api.md` (old call → new `PasswordKey`/options equivalent). Right now the doc only describes the new shape, so an upgrading user gets compile errors with no pointer to the fix.

Activities/Cryptography/UiPath.Cryptography.Activities.API/Services/CryptographyService.cs
```diff
@@ -11,708 +11,351 @@ namespace UiPath.Cryptography.Activities.API
...
-            byte[] keyBytes = SecureStringToBytes(key, encoding);
-            try
-            {
-                return DecryptText(input, algorithm, keyBytes, encoding);
-            }
-            finally
-            {
-                Array.Clear(keyBytes, 0, keyBytes.Length);
-            }
+            byte[] cipher = EncryptBytes(Encoding.UTF8.GetBytes(input), algorithm, options);
```
LiviuPonova: `EncryptText`/`DecryptText` now hardcode UTF-8 (`Encoding.UTF8.GetBytes`/`GetString`), and neither `SymmetricEncryptOptions` nor `SymmetricDecryptOptions` can express a different plaintext encoding. The old overloads took an `Encoding`, so ciphertext produced from non-UTF-8 plaintext via the pre-PR coded API will now decrypt to mojibake — silently, since the call compiles fine after migration.

(The XAML activities are unaffected — `EncryptText.cs` still applies the user-selected encoding to the plaintext — so this is scoped to coded-workflow callers.)

Suggest either an optional `Encoding` on the options classes, or at minimum a doc callout that the text APIs are UTF-8-only and non-UTF-8 callers should drop to the `*Bytes` APIs.

Activities/Cryptography/UiPath.Cryptography.Activities/EncryptText.cs
```diff
@@ -199,21 +228,39 @@ private string ExecuteSymmetricEncrypt(CodeActivityContext context, string input
             var keySecureString = KeySecureString.Get(context);
             var keyEncoding = Encoding.Get(context);
             var keyEncodingString = KeyEncodingString.Get(context);
+            var ivString = Iv?.Get(context);
+            var iterations = KdfIterations?.Get(context) ?? 0;
+
+            SymmetricInteropHelper.ValidateInteropSettings(Algorithm, Format, KeyFormat, ivString, iterations, null);
```
LiviuPonova: This validate → parse key/IV → re-validate(Raw) → dispatch → `finally { ClearKeyBytes }` sequence is repeated nearly verbatim in all four activities (DecryptText ~221-254, EncryptFile ~262-277, DecryptFile ~284-302) — only the payload source and direction differ.

Normally I'd let 4x duplication slide, but the duplicated block includes the security-sensitive key-zeroing `finally`: a future fix applied to three of the four copies silently leaves the fourth leaking key bytes, and each file looks locally correct in review. Consider extracting a shared helper (e.g. on `SymmetricInteropHelper`: take direction + settings + payload `byte[]`, own the parse/validate/clear lifecycle) so the lifecycle lives in one place.

Activities/Cryptography/UiPath.Cryptography.Activities.API/Services/CryptographyService.cs
```diff
@@ -11,708 +11,351 @@ namespace UiPath.Cryptography.Activities.API
...
+            byte[] keyBytes = options.Key.KeyBytes;
             try
             {
-                return EncryptText(input, algorithm, keyBytes, encoding);
+                return SymmetricInteropHelper.DispatchEncrypt(algorithm, options.Format, options.KdfIterations, keyBytes, options.IV, input);
             }
             finally
             {
-                Array.Clear(keyBytes, 0, keyBytes.Length);
+                options.Key.ReleaseMaterialisedBytes(keyBytes);
```
LiviuPonova: The key-zeroing contract here relies on a hand-written `try/finally { Key.ReleaseMaterialisedBytes(keyBytes) }` at every `KeyBytes` consumption site — six in this file. Nothing stops the next method added to this service from forgetting the `finally`, which would silently reintroduce exactly the unzeroed-password-bytes issue this PR fixes (P2), and neither the compiler nor tests would catch it.

Consider a scoped accessor on `CryptoKey`, e.g. `T UseKeyBytes<T>(Func<byte[], T> use)` that materialises, invokes, and releases in one place. The six call sites become one-liners and the cleanup can't be bypassed by construction.

LiviuPonova (commented):
Reviewed with focus on API compatibility and key-material handling. Four comments inline: two on the coded-API contract (breaking overload removal, forced UTF-8) and two on the key-zeroing pattern (duplicated across the four activities, and enforceable-by-construction in the service). The crypto core itself held up well under review — wire-format offsets, Raw round-trips, and KDF usage all checked out.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
```diff
@@ -124,6 +129,47 @@ protected void ConfigureAlgorithmAndKeyProperties(ref int orderIndex)
             _encodingDataSource.Data = EncodingHelpers.GetAvailableEncodings();
         }
 
+        /// <summary>
+        /// Configures the third-party-compatibility properties (Format, KeyFormat, Iv, KdfIterations).
+        /// Format is visible by default; the others are hidden until <see cref="ApplyInteropVisibility"/>
+        /// reveals them based on Format/Algorithm.
+        /// </summary>
+        protected void ConfigureInteropProperties(ref int orderIndex)
```
LiviuPonova, line 137: The new interop configuration added here is duplicated near-verbatim in `DecryptCryptoViewModelBase`: `ConfigureInteropProperties`, `FormatChanged_Action`, `ApplyInteropVisibility`, and the Owasp2026 `KdfIterations` label switch are essentially identical on both sides (docstrings included) — the only real deltas are the encrypt-side `Iv` property and one `PublicKey` visibility condition. That's ~90 new duplicated lines on top of the algorithm/key configuration that was already mirrored between the two classes pre-PR.

Since both classes inherit `DesignPropertiesViewModel`, a shared intermediate base (or a helper both call into) could own the common format/visibility/label logic, with the encrypt/decrypt deltas as small overrides. Not blocking — but the PR explicitly anticipates future wire formats (Owasp2030), and each one will mean editing this logic twice, where updating one side and missing the other silently diverges the Encrypt and Decrypt designers.

Activities/Cryptography/UiPath.Cryptography.Activities/Properties/UiPath.Cryptography.Activities.de.resx
```diff
@@ -1012,75 +1012,75 @@
     <value>Die Passphrase für den privaten PGP-Schlüssel, der zum Signieren verwendet wird.</value>
     <comment>Property description</comment>
   </data>
-  <data name="Activity_PgpGenerateKeys_Name" xml:space="preserve">
-    <value>Generierte PGP-Schlüssel</value>
+  <data name="Activity_PgpGenerateKeyPair_Name" xml:space="preserve">
+    <value>PGP Generate Keys</value>
```
LiviuPonova: Please revert the changes to all 13 localized resx files (`*.de.resx`, `*.fr.resx`, `*.ja.resx`, …) and keep only the neutral `UiPath.Cryptography.Activities.resx` edits. The localized files are owned by the Localization pipeline (see the periodic `chore(l10n): sync from Localization` PRs, e.g. #567) — hand edits risk conflicting with or being overwritten by the next sync.

It also avoids a few accidental translation regressions this PR introduces: this line replaces the German `Generierte PGP-Schlüssel` with English, and the same happens in ro (`Generați chei PGP`, `Fișier PGP Clearsign`) and tr (`PGP Anahtar Oluştur`, `PGP Şeffaf İmza Dosyası`).

After the revert, the renamed keys (`Activity_PgpGenerateKeyPair_*`, `Activity_PgpClearSignFile_*`) will fall back to English in non-English locales until the next l10n sync picks them up from the neutral file — that's the expected interim state for renames.

LiviuPonova (changes requested):

Activities/Cryptography/UiPath.Cryptography.Activities.Tests/PgpClearSignFileBranchTests.cs
```diff
@@ -0,0 +1,143 @@
...
+        [Fact]
+        public void PgpClearSignFile_TypeName_IsPascalCase()
+        {
+            Type t = typeof(PgpClearSignFile);
+            t.Name.ShouldBe("PgpClearSignFile");
+            t.FullName.ShouldBe("UiPath.Cryptography.Activities.PgpClearSignFile");
+
+            // No stale camelCase type lingering in the assembly.
+            Type stale = t.Assembly.GetType("UiPath.Cryptography.Activities.PgpClearsignFile", throwOnError: false);
+            stale.ShouldBeNull("the old PgpClearsignFile (camelCase) name should not exist after the rename");
```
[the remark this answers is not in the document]
alexandru-petre: The Pgp clear-sign activities (both the new `PgpClearSignFile` and its in-repo predecessor) have never shipped in a published `UiPath.Cryptography.Activities` package, so there is no installed base whose persisted XAML could reference `UiPath.Cryptography.Activities.PgpClearsignFile`. The rename is a pre-publication correction, not a breaking change. The test pinning the absence of the camelCase type stays.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
alexandru-petre: Agreed this is worth doing, and not blocking. Tracked as STUD=80470  — the shared base needs a manual Studio designer smoke-test across all four concrete activities (`EncryptText`, `EncryptFile`, `DecryptText`, `DecryptFile`) because the repo has no ViewModel unit tests, so the verification surface is fragile and warrants its own focused PR.

Activities/Cryptography/UiPath.Cryptography.Activities.API/Services/ICryptographyService.cs
alexandru-petre: Migration section added to `coded-api.md` (`Migrating from the prior coded API`) — covers the one-to-one old→new mapping for every symmetric / keyed-hash / PGP overload, the plaintext-encoding behaviour (now on `SymmetricEncrypt/DecryptOptions.TextEncoding`, default UTF-8), and the path-based-`PgpGenerateKeys` migration to the in-memory `PgpKeyPair` + `.Save(path)` pattern. Resolving this thread; the explicit owner sign-off is being handled separately.

Activities/Cryptography/UiPath.Cryptography.Activities.Tests/PgpClearSignFileBranchTests.cs
```diff
@@ -0,0 +1,143 @@
...
+            t.GetProperty(nameof(PgpClearSignFile.PassphraseSecureString)).ShouldNotBeNull();
+            t.GetProperty(nameof(PgpClearSignFile.OutputFilePath)).ShouldNotBeNull();
+            t.GetProperty(nameof(PgpClearSignFile.ClearSignedFile)).ShouldNotBeNull();
+        }
+
+        // Pre-cancelled token — the async file resolution awaits the token and must throw
+        // OperationCanceledException before reaching the synchronous sign helper.
+        [Fact]
+        public void PgpClearSignFile_PreCancelledToken_ThrowsOperationCancelled()
+        {
```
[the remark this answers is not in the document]
alexandru-petre: Already addressed in 61be28a — the misleading PreCancelledToken test was rewritten as `PgpClearSignFile_DefaultContinueOnError_MissingInput_Throws` (drops the CancellationTokenSource, asserts the actual missing-input failure path).

Activities/Cryptography/UiPath.Cryptography.Activities.Tests/PgpClearSignFileBranchTests.cs
```diff
@@ -0,0 +1,143 @@
...
+                    Passphrase = new InArgument<string>(Passphrase),
+                    OutputFilePath = new InArgument<string>(outputPath),
+                    Overwrite = true,
+                };
+
+                // Reach the async ExecuteAsync directly with a pre-cancelled token. The activity's
+                // PgpFileResolver.ResolveAsync awaits the token internally — pre-cancelling
+                // it makes the first await observe cancellation and throw.
+                using var cts = new CancellationTokenSource();
+                cts.Cancel();
```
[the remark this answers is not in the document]
alexandru-petre: Already addressed in 61be28a — the unused `CancellationTokenSource` and its dead-code surroundings were removed when the test was rewritten.

Activities/Cryptography/UiPath.Cryptography.Activities/Properties/UiPath.Cryptography.Activities.tr.resx
```diff
@@ -1012,75 +1012,75 @@
     <value>İmzalamak için kullanılan PGP özel anahtarının parolası.</value>
     <comment>Property description</comment>
   </data>
-  <data name="Activity_PgpGenerateKeys_Name" xml:space="preserve">
-    <value>PGP Anahtar Oluştur</value>
+  <data name="Activity_PgpGenerateKeyPair_Name" xml:space="preserve">
+    <value>PGP Generate Keys</value>
```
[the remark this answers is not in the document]
alexandru-petre: Already addressed in 6102921 — the 12 localized resx files were reverted to develop; the Turkish PgpGenerateKeys translation is restored.

Activities/Cryptography/UiPath.Cryptography.Activities/Properties/UiPath.Cryptography.Activities.tr.resx
```diff
@@ -1148,67 +1148,67 @@
     <value>Dosya kaynağı olarak imzalanan dosya.</value>
     <comment>Property description</comment>
   </data>
-  <data name="Activity_PgpClearsignFile_Name" xml:space="preserve">
-    <value>PGP Şeffaf İmza Dosyası</value>
+  <data name="Activity_PgpClearSignFile_Name" xml:space="preserve">
+    <value>PGP Clear Sign File</value>
```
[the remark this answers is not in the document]
alexandru-petre: Already addressed in 6102921 — Turkish PgpClearsignFile translation restored as part of the localized-resx revert.

Activities/Cryptography/UiPath.Cryptography.Activities/Properties/UiPath.Cryptography.Activities.de.resx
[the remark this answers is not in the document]
alexandru-petre: Already addressed in 6102921 — German PgpGenerateKeys translation restored as part of the localized-resx revert.

Activities/Cryptography/UiPath.Cryptography.Activities.API/Models/SymmetricEncryptOptions.cs
```diff
@@ -0,0 +1,66 @@
...
+        /// <b>⚠ NEVER reuse the same (Key, IV) pair across encryptions.</b> Doing so destroys
+        /// confidentiality (CTR keystream reuse) and — under AEAD modes (AES-GCM,
+        /// ChaCha20-Poly1305) — additionally lets an attacker recover the authentication key
+        /// and forge arbitrary messages under that key. Prefer leaving <paramref name="iv"/>
+        /// null so a fresh random IV is generated per call; only supply an explicit IV when a
+        /// third-party protocol mandates it, and ensure your producer guarantees uniqueness.
+        /// </para>
+        /// </param>
+        public static SymmetricEncryptOptions Raw(RawKey key, byte[] iv = null) =>
+            new() { Key = key, Format = SymmetricWireFormat.Raw, IV = iv };
```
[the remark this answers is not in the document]
alexandru-petre: Addressed in 4e20f5b — `SymmetricEncryptOptions.Raw(...)` now `Buffer.BlockCopy`s the supplied iv into a new array before storing, matching the existing precedent on `RawKey.FromBytes` / `PgpPublicKey.FromBytes` / `PgpPrivateKey.FromBytes`. Added a regression test (`SymmetricEncryptOptions_Raw_DefensiveCopiesIv_CallerMutationDoesNotAffectEncryption`) that `Array.Clear`s the caller’s iv buffer between two encryptions reusing the same options object and asserts byte-equal ciphertexts — fails on the pre-fix code, passes after.

Activities/Cryptography/UiPath.Cryptography.Activities.API/Services/CryptographyService.cs
```diff
@@ -11,708 +11,333 @@ namespace UiPath.Cryptography.Activities.API
...
-            byte[] keyBytes = SecureStringToBytes(key, encoding);
-            try
-            {
-                return KeyedHashFile(filePath, algorithm, keyBytes);
-            }
-            finally
-            {
-                Array.Clear(keyBytes, 0, keyBytes.Length);
-            }
+            return key.UseKeyBytes(keyBytes => ComputeHashHex(algorithm, Encoding.UTF8.GetBytes(input), keyBytes));
```
LiviuPonova: **`KeyedHashText`/`KeyedHashFile` now hardcode UTF-8 for the input — confirm this is intentional?**

The old overloads applied the caller's `encoding` to the input (`byte[] inputBytes = encoding.GetBytes(input)`); the new signature drops the parameter and always uses `Encoding.UTF8.GetBytes(input)`.

For ASCII input this is identical, but for **non-ASCII** input a caller who previously passed a non-UTF-8 encoding (e.g. UTF-16) will now get a **different HMAC**, silently — so a coded workflow verifying digests computed against the old API would see mismatches.

UTF-8 is a fine default and I suspect this is deliberate, but since it's a behavior change in a public API, could we (a) confirm it's intended and (b) add a line to the migration notes? Non-blocking.

LiviuPonova (approved):
Approving.

Reviewed the diff across correctness, removed-behavior, cross-file impact, and cleanup/altitude angles. No merge-blocking issues found:

- Classic wire format stays byte-stable and is pinned by hex-literal stability fixtures; existing ciphertext keeps decrypting.
- Backward compat holds — `Format` defaults to `Classic` and `KdfIterations` to `0`, routing old workflows through the original code paths.
- Strong test coverage (858 tests + interop round-trips).

Refuted several candidate issues on inspection (compile errors from "missing usings" — the enums resolve via the enclosing namespace; double-validation — guarded by `format == Raw`; un-encapsulated key cleanup — it's centralized in `CryptoKey.UseKeyBytes`).

One non-blocking question left inline: `KeyedHashText`/`KeyedHashFile` now hardcode UTF-8 for the input where the old overloads honored the caller's encoding — please confirm it's intentional and add a line to the migration notes. Doesn't gate the merge.

Note: this is a functional/design review, not a dedicated cryptographic security audit of the primitives.

Activities/Cryptography/UiPath.Cryptography.Activities.API/Services/CryptographyService.cs
alexandru-petre: Good catch — confirmed intentional UTF-8 default, but the silent behaviour change for non-UTF-8 callers warrants more than a doc note. Addressed in 6a83c21: added an optional trailing `Encoding encoding = null` parameter to `KeyedHashText` (interface + impl) so callers can pass the same encoding they used pre-PR and get identical digests; default UTF-8 preserves the current behaviour. Added a regression test (`KeyedHashText_NonUtf8Encoding_ProducesDifferentDigestThanUtf8`) that computes the same input under UTF-8 and UTF-16 and asserts the digests differ, plus pins UTF-8 as the default.

`KeyedHashFile` intentionally not changed: file contents are hashed byte-for-byte, so there's no plaintext-encoding step. The prior overload's `Encoding` parameter only scrubbed the *password* — that role moved to `PasswordKey.FromPassword(string, Encoding)` and is already supported. The migration table in `coded-api.md` now covers both `KeyedHashText` (encoding passed both to `PasswordKey.FromPassword` and as the trailing parameter) and `KeyedHashFile` (only to `PasswordKey.FromPassword`).

PR #572, opened 2026-06-18
fix(cryptography): decouple plaintext encoding + make public Encoding/PlaintextEncoding authoritative over their proxies [STUD-80530][STUD-80559]

alexandru-petre:
## Problem

Two related defects in the `EncryptText` / `DecryptText` (and sibling) encoding handling:

**1. `Encoding` dual-role (STUD-80530).** The activities read a single `Encoding` property and used it for **two unrelated concerns**: parsing the password/key bytes *and* converting the plaintext to bytes. For the `Raw` / `OpenSslEnc` interop formats (STUD-80231), a non-UTF-8 codepage produced ciphertext whose plaintext a Python `cryptography` / Java `javax.crypto` / `openssl enc -d` consumer decoded as mojibake. UiPath↔UiPath round-trips hid it.

**2. Public encoding properties masked by their proxies (STUD-80530 P1 + STUD-80559).** Each encoding is a *paired* input: a public `InArgument<Encoding>` plus a hidden `KeyEncodingString` / `PlaintextEncodingString` code-page proxy for the modern designer. They were merged by `EncodingHelpers.KeyEncodingOrString`, which gave the **string proxy precedence whenever non-null**. Because every constructor defaults the proxy to UTF-8, the proxy was *never* null — so the public `Encoding` / `PlaintextEncoding` InArgument was **silently ignored** (e.g. `new EncryptText { PlaintextEncoding = Encoding.Unicode }` resolved to UTF-8) unless the caller also cleared the hidden proxy. This affects **Legacy designer / hand-authored XAML / programmatic** use; modern Windows and cross-platform designers bind only the proxy, so they were never affected.

## Analysis

The coded API surface is **already correct** (`PasswordKey._encoding` is separate from `SymmetricEncryptOptions.TextEncoding`). This is an **activity-layer-only** fix. The masking precedence dates to commit `c0e892d` (STUD-64242), which flipped the merge to "string wins" while fixing a cross-platform `[RequiredArgument]` design-time error; the runtime flip was incidental and, combined with the always-present UTF-8 constructor default, is what hid the public properties.

## Solution

**Decouple (STUD-80530):** add a `PlaintextEncoding` property (paired `InArgument<Encoding>` + `PlaintextEncodingString` proxy) governing plaintext↔bytes, defaulting to UTF-8. The existing `Encoding` keeps its password/key role. `EncryptFile` / `DecryptFile` are unaffected (raw bytes).

**Fix precedence (STUD-80530 P1 + STUD-80559):** flip `EncodingHelpers.KeyEncodingOrString` to prefer the explicit `InArgument<Encoding>` when bound, falling back to the code-page proxy otherwise (and returning `null` when neither is set, so the caller applies its UTF-8 default):

```csharp
if (keyEncoding != null) return keyEncoding;                 // typed InArgument (Legacy/XAML/programmatic) wins
if (keyEncodingString != null) return /* parse code page */; // modern-designer proxy + UTF-8 ctor default
return null;
```

This one helper change makes the public property authoritative across **all six activities that share it** — `EncryptText`, `DecryptText`, `EncryptFile`, `DecryptFile`, `KeyedHashText`, `KeyedHashFile` — for both the key/password and plaintext encodings.

**Cross-platform safety:** modern Windows + cross-platform bind only the proxy, so the typed InArgument is `null` at runtime and the proxy still governs — byte-stable, no regression. `[RequiredArgument]` stays removed, so STUD-64242's design-time fix is preserved.

## Commits

1. `decouple plaintext encoding from key encoding` — adds `PlaintextEncoding` (STUD-80530).
2. `honor public PlaintextEncoding over its proxy` — plaintext precedence fix (STUD-80530 P1 review feedback).
3. `honor public key Encoding over its proxy across symmetric + keyed-hash activities` — generalizes the precedence fix to the key/password pair via the shared helper and consolidates the now-identical `PlaintextEncodingOrString` away (STUD-80559).

## Testing

- `dotnet build Activities/Activities.Cryptography.sln` — 0 errors, no new warnings.
- `dotnet test Activities/Activities.Cryptography.sln` — **776 + 123 (API) passed, 0 failed, 0 skipped**; `WireFormatStabilityTests` byte-identical.
- Regression tests (each confirmed to **fail on pre-fix code**):
  - `OpenSslEnc_AesCbc_PlaintextEncodingUtf16_PreservedForExternalConsumer` — UTF-16 plaintext decoded by a BCL counterpart.
  - `OpenSslEnc_AesCbc_PublicPlaintextEncodingWinsOverProxyDefault` — public `PlaintextEncoding` wins with the proxy left at its UTF-8 default.
  - `OpenSslEnc_AesCbc_PublicKeyEncodingWinsOverProxyDefault` — public key `Encoding` (UTF-16) governs key derivation with the proxy left at default.
  - `KeyEncodingOrString_TypedEncoding_WinsOverProxyString` — unit-level precedence guard.
  - `OpenSslEnc_AesCbc_KeyEncodingDoesNotAffectPlaintextBytes` — key and plaintext encoding stay decoupled.

## Caveats

- Localized `*.{lang}.resx` variants are intentionally not pre-created; they follow via the standard Localization sync.
- The precedence fix is a behavior change on long-shipped properties, but only for Legacy/XAML/programmatic callers that set a typed encoding whose bytes differ from UTF-8; modern designer workflows are byte-stable.

Closes [STUD-80530](https://uipath.atlassian.net/browse/STUD-80530) and [STUD-80559](https://uipath.atlassian.net/browse/STUD-80559).

[STUD-80530]: https://uipath.atlassian.net/browse/STUD-80530?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
@@ -140,7 +140,6 @@ namespace UiPath.Cryptography.Activities.NetCore.ViewModels
         {
             var dataSource = EncodingHelpers.ConfigureEncodingDataSource();
             encodingProperty.IsPrincipal = false;
-            encodingProperty.IsVisible = true;
             encodingProperty.OrderIndex = orderIndex++;
             encodingProperty.Category = Resources.Input;
             encodingProperty.DataSource = dataSource;
--- a/Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
@@ -132,6 +132,23 @@ namespace UiPath.Cryptography.Activities.NetCore.ViewModels
         }
 
         /// <summary>
+        /// Configures a code-page encoding dropdown on the supplied design property, mirroring the
+        /// <see cref="KeyEncodingString"/> setup. Used by Text activities to surface their plaintext
+        /// encoding; File activities operate on raw bytes and do not call this.
+        /// </summary>
+        protected void ConfigureEncodingDropdown(DesignInArgument<string> encodingProperty, ref int orderIndex)
+        {
+            var dataSource = EncodingHelpers.ConfigureEncodingDataSource();
+            encodingProperty.IsPrincipal = false;
+            encodingProperty.IsVisible = true;
+            encodingProperty.OrderIndex = orderIndex++;
+            encodingProperty.Category = Resources.Input;
+            encodingProperty.DataSource = dataSource;
+            encodingProperty.Widget = new DefaultWidget { Type = ViewModelWidgetType.Dropdown, Metadata = new Dictionary<string, string>() };
+            dataSource.Data = EncodingHelpers.GetAvailableEncodings();
+        }
+
+        /// <summary>
         /// Configures the third-party-compatibility properties (Format, KeyFormat, Iv, KdfIterations).
         /// Format is visible by default; the others are hidden until <see cref="ApplyInteropVisibility"/>
         /// reveals them based on Format/Algorithm.
@@ -316,7 +333,14 @@ namespace UiPath.Cryptography.Activities.NetCore.ViewModels
             PrivateKeyFilePath.IsPrincipal = isPgp && SignData.Value;
             ApplyPassphraseVisibility();
             ApplyInteropVisibility();
+            OnAlgorithmChanged(isPgp);
         }
+
+        /// <summary>
+        /// Hook invoked at the end of <see cref="AlgorithmChanged_Action"/> so derived ViewModels can
+        /// react to the active algorithm (e.g. hide Text-only properties when PGP is selected).
+        /// </summary>
+        protected virtual void OnAlgorithmChanged(bool isPgp) { }
 
         private void FormatChanged_Action()
         {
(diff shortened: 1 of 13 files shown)


Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptTextViewModel.cs
```diff
@@ -23,6 +23,7 @@ public EncryptTextViewModel(IDesignServices services) : base(services)
         }
 
         public DesignInArgument<string> Input { get; set; } = new DesignInArgument<string>();
+        public DesignInArgument<string> PlaintextEncodingString { get; set; } = new() { Name = nameof(PlaintextEncodingString) };
```
lucarachiteanu:
```suggestion
        public DesignInArgument<string> PlaintextEncodingString { get; set; } = new();
```

The base ViewModel will simply override this value.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
```diff
@@ -131,6 +131,23 @@ protected void ConfigureAlgorithmAndKeyProperties(ref int orderIndex)
...
+        /// <summary>
+        /// Configures a code-page encoding dropdown on the supplied design property, mirroring the
+        /// <see cref="KeyEncodingString"/> setup. Used by Text activities to surface their plaintext
+        /// encoding; File activities operate on raw bytes and do not call this.
+        /// </summary>
+        protected void ConfigureEncodingDropdown(DesignInArgument<string> encodingProperty, ref int orderIndex)
+        {
+            var dataSource = EncodingHelpers.ConfigureEncodingDataSource();
+            encodingProperty.IsPrincipal = false;
+            encodingProperty.IsVisible = true;
```
lucarachiteanu:
```suggestion
```
Principal is false by default, and Visible is true by default.

lucarachiteanu (approved):

PR #575, opened 2026-06-19
fix(crypto): Gate IV nonce-reuse warning to Raw format [STUD-80532]

alexandru-petre:
## Context
`EncryptText` and `EncryptFile` (UiPath.Cryptography.Activities) are the
symmetric/PGP encryption activities in the Cryptography package. At design time
they raise validation warnings from `CacheMetadata` that Studio surfaces on the
activity. One such warning, added under STUD=64429 / STUD=80231 alongside the
`Raw` wire format and the explicit `Iv` property, cautions against reusing a
`(Key, IV)` pair.

## Problem Statement
The IV nonce-reuse warning fired far too eagerly. Its only guard was
`if (Iv != null)`, with no regard for the selected wire `Format` or `Algorithm`.
An IV is consumed *only* by the `Raw` format on symmetric algorithms — the
runtime rejects an IV on `Classic` / `Owasp2026` / `OpenSslEnc`
(`SymmetricInteropHelper`, `Validation_Iv_OnlyForRaw`) and PGP ignores it
entirely. So an author who set an `Iv` and then switched away from `Raw` kept
seeing the alarming "NEVER reuse (Key, IV)" warning in contexts where no IV is
used — desensitizing them before a real Raw + explicit-IV reuse risk.

## Behavior Before This PR
An author sets `Iv` on `EncryptText`, then sets `Format = Classic` (or picks
`Algorithm = PGP`). The design-time warning "NEVER reuse the same (Key, IV)
pair…" still shows, even though that configuration never consumes the IV at
runtime.

## Behavior After This PR
Same scenario: with `Format = Classic` (or any non-Raw format), or with
`Algorithm = PGP`, no IV warning is shown. The warning appears only for the one
configuration that actually consumes an explicit IV — symmetric algorithm +
`Format = Raw` — exactly mirroring the runtime's IV-consumption path.

## Considered Use Cases
- Symmetric + `Raw` + explicit IV → warns (the real reuse risk).
- Symmetric + `Classic`/`Owasp2026`/`OpenSslEnc` + IV → no warning (runtime rejects the IV).
- `Raw` + empty IV → no warning (legitimate "generate a random IV per call").
- `PGP` + `Raw` + IV (hand-edited/persisted XAML) → no warning (PGP ignores Format/Iv).

## Implementation
The `CacheMetadata` guard becomes
`Iv != null && Format == SymmetricWireFormat.Raw && Algorithm != EncryptionAlgorithm.PGP`
in both activities. `Format` and `Algorithm` are plain properties (not
`InArgument`), so they are readable at design time. The wrapper-vs-value aspect
of `Iv != null` (a touched-but-empty IV field) is intentionally left as-is per
the ticket — it can't be reliably evaluated from `CacheMetadata`, and "Raw +
empty IV" is a valid configuration.

Test assertions for this warning were standardized on exact resource equality
(`w.Message == Resources.Iv_NonceReuseWarning`) instead of brittle substring
matching, via an aliased `using` to disambiguate the production `Resources`
class from the test project's own `Resources` namespace.

## How to Test
- `dotnet test Activities/Activities.Cryptography.sln --filter "FullyQualifiedName~CacheMetadataWarningTests"` — 18/18 pass.
- The negative tests fail if the `Format == Raw` / `Algorithm != PGP` gate is removed (verified), confirming they're genuine reproducers.
- Full pack regression: `dotnet test Activities/Activities.Cryptography.sln` stays green.

### DIFF
--- a/Activities/Cryptography/UiPath.Cryptography.Activities.Tests/CacheMetadataWarningTests.cs
+++ b/Activities/Cryptography/UiPath.Cryptography.Activities.Tests/CacheMetadataWarningTests.cs
@@ -5,6 +5,7 @@ using System.Linq;
 using Shouldly;
 using UiPath.Cryptography.Enums;
 using Xunit;
+using CryptoRes = UiPath.Cryptography.Activities.Properties.Resources;
 
 #pragma warning disable CS0618 // tests intentionally use obsolete algorithms to fire FIPS warnings.
 
@@ -48,8 +49,8 @@ namespace UiPath.Cryptography.Activities.Tests
         }
 
         // Iv set on EncryptText with Format == Raw → IV nonce-reuse warning. The IV is only
-        // consumed by the Raw wire format, so the warning is gated on it. The warning text
-        // mentions "(Key, IV) pair" so the user can self-serve.
+        // consumed by the Raw wire format, so the warning is gated on it. Matched by exact
+        // resource equality so a reword of the message can't silently weaken the assertion.
         [Fact]
         public void EncryptText_WithExplicitIv_EmitsNonceReuseWarning()
         {
@@ -61,7 +62,7 @@ namespace UiPath.Cryptography.Activities.Tests
             };
             ValidationError[] warnings = ValidateAndGetWarnings(activity);
 
-            warnings.Any(w => w.Message.Contains("(Key, IV) pair", StringComparison.Ordinal) || w.Message.Contains("explicit IV", StringComparison.Ordinal)).ShouldBeTrue(
+            warnings.Any(w => w.Message == CryptoRes.Iv_NonceReuseWarning).ShouldBeTrue(
                 $"expected an IV nonce-reuse warning, got: {Format(warnings)}");
         }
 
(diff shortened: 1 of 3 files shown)


AlexMarinescuUiPath (approved):
## Review: Approve (minor nits, non-blocking)

Sound, well-scoped, and backward compatible. The design-time gate now mirrors
the runtime's IV-consumption path, both activities stay aligned, and the tests
were upgraded to exact resource-equality matching. No blocking findings.

### Correctness (verified against runtime)
- The new guard matches the runtime exactly: `SymmetricInteropHelper.ValidateInteropSettings`
  hard-rejects an IV for any non-Raw format (`Validation_Iv_OnlyForRaw`), and PGP
  routes through the PGP path, bypassing the symmetric helper. Gating on
  `Format == Raw && Algorithm != PGP` is therefore the precise design-time mirror.
- `Format` and `Algorithm` are plain CLR properties (not InArgument), so they are
  readable in `CacheMetadata` at design time.
- `Format` defaults to `Classic`, so default activities do not warn unless Raw is
  explicitly chosen.
- No public-contract change, no runtime behavior change, no resx change. Fully
  backward compatible. Decrypt activities correctly untouched.

### Strengths
- Both activities changed identically and pinned by parallel tests.
- Assertions moved from brittle substring matching to exact resource equality
  (`== Resources.Iv_NonceReuseWarning`), so a message reword cannot silently
  weaken a test.

### Minor findings (non-blocking)
1. `Iv != null` reflects whether the argument is bound, not whether it holds a
   value. So "Raw + empty IV -> no warning" holds only for a never-bound IV; a
   Raw config where the author typed then cleared the IV still warns. This is an
   acknowledged CacheMetadata limitation; suggest a one-line caveat in the PR
   description so the use-case table is not read as absolute.
2. Test gap: the `Iv != null` clause is not pinned under `Format == Raw`. The
   "no IV" test uses the default `Format = Classic`, so a mutation dropping
   `Iv != null` would survive. Suggest adding `Raw + no IV -> no warning` for
   both EncryptText and EncryptFile.
3. UX (out of scope): non-Raw + IV now shows nothing at design time, then throws
   at runtime. Consider a follow-up ticket for a targeted "IV is only used with
   Raw format" hint.

PR #584, opened 2026-06-26
FTP: make SFTP Delete recursive for non-empty directories [STUD-80569]

viogroza:
## Problem

Jira: [STUD-80569](https://uipath.atlassian.net/browse/STUD-80569)

The `Delete` activity fails when the target is a **non-empty directory over SFTP**. `SftpSession.Delete`/`DeleteAsync` called `SftpClient.Delete(path)`, which issues a single `SSH_FXP_RMDIR`. The SFTP protocol only permits `RMDIR` on an **empty** directory, so the server returns a failure and SSH.NET throws. SSH.NET ships no built-in recursive delete.

The FTP path was already fine — FluentFTP's `DeleteDirectory(path)` is recursive by default.

## Fix

Replace the one-shot delete in `SftpSession` with a depth-first recursive walk used by both the sync `Delete` and async `DeleteAsync`:

- Removes a directory's contents before removing the directory itself.
- **Symbolic links are unlinked, not followed** (`IsDirectory && !IsSymbolicLink`), mirroring `rm -rf` and avoiding loops/escaping the tree.
- Skips `.` and `..` entries.
- Honors the `CancellationToken` throughout the walk (the previous one-shot `Task.Run` did not).
- Preserves the prior "path not found" behavior (missing path surfaces the same `SftpPathNotFoundException`).

This brings SFTP in line with FTP so both protocols delete non-empty folders consistently. No `rm -rf`/shell dependency — the walk stays within the SFTP subsystem, so it remains portable across SFTP-only and non-POSIX servers and has no command-injection surface.

Added `ClientGet`/`ClientListDirectory` `protected internal virtual` seams (alongside the existing `ClientExists`) so the walk is unit-testable without a live server.

## Behavior change

Deleting a non-empty directory over **SFTP** previously threw; it now succeeds recursively (matching FTP). Flagging per the repo's "breaking changes require discussion" convention. Kept as silent-recursive rather than gated behind a new opt-in `Recursive` property — happy to switch to opt-in if preferred.

## Tests

6 new `SftpSessionTests` (all green; 21 total pass, builds clean on net6.0 and net6.0-windows):
- depth-first deletion order (children before parent, target last)
- single-file delete without listing
- symlink not recursed into
- `.`/`..` skipped
- cancellation throws and deletes nothing
- sync `Delete` path

[STUD-80569]: https://uipath.atlassian.net/browse/STUD-80569?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/Activities/FTP/UiPath.FTP.Tests/SessionsTests/SftpSessionTests.cs
+++ b/Activities/FTP/UiPath.FTP.Tests/SessionsTests/SftpSessionTests.cs
@@ -1,5 +1,7 @@
+using Moq;
 using Renci.SshNet;
 using Renci.SshNet.Common;
+using Renci.SshNet.Sftp;
 using System;
 using System.Collections.Generic;
 using System.Linq;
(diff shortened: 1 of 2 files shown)


AlexMarinescuUiPath (approved):
looks correct and safe to merge from a code-quality and regression-risk perspective.

PR #588, opened 2026-07-08
feat(cryptography): unified-canvas UX + PGP private-key resource toggle [STUD-80718]

alexandru-petre:
## Context

`UiPath.Cryptography.Activities` is the Cryptography activity package (Encrypt/Decrypt Text & File, Hash Text & File, and the PGP activities). Each activity has a runtime class (`CodeActivity`) and a design-time `DesignPropertiesViewModel` that drives the **unified canvas** — the property panel and card shown in Studio. Property placement there is controlled by each property's ViewModel `Category`/`IsPrincipal`/`OrderIndex`, kept in sync with the runtime `[LocalizedCategory]` attributes.

This PR implements the follow-up fixes from the **STUD-80718** unified-canvas adoption review, plus polish items surfaced while doing so.

## Problem Statement

The review found the Encrypt/Decrypt/Hash activities didn't fully follow the unified-canvas guideline: interop and encoding knobs were scattered across the primary Input section and Options; outputs weren't consistently ordered last; `Overwrite` sat outside Options on the file activities; and the PGP **private key** could only be given as a file path — unlike its sibling public key and the PgpSign activities, it had no path↔resource toggle. The IV nonce-reuse risk was also only surfaced after a value was entered. These affect automation authors configuring these activities in Studio.

## Behavior Before This PR

Opening **Encrypt Text** in Studio: Text encoding sat under **Input**, Key encoding under **Options**, and the third-party knobs (Format, KeyFormat, IV, KDF iterations, AES key size) under **Input** alongside the primary fields; the **Result** output could render before the Options tail. On **Encrypt/Decrypt File**, **Overwrite** surfaced on the collapsed card rather than in Options. For PGP, the private key (`PrivateKeyFilePath`) was a plain string only — it could not be picked from project resources / a Storage Bucket. The IV field's nonce-reuse warning only appeared as a design-time validation once an explicit IV was set.

## Behavior After This PR

Encrypt/Decrypt Text sections read **Input → Advanced → Encoding → Options → Output**: both encodings live in a dedicated **Encoding** section, the interop knobs in **Advanced**, and the output is ordered last. **Overwrite** renders in **Options** on Encrypt/Decrypt File (consistent with the PGP file activities). The PGP private key now offers a **path↔resource toggle** (`PrivateKeyFile`, `IResource`) via a designer menu action, mirroring `PublicKeyFile` and PgpSign; at runtime the file path is used first, falling back to the resource only when the path is empty. The IV property description now carries a short `WARNING:` about never reusing a (Key, IV) pair, in addition to the existing design-time validation.

## Considered Use Cases

These all concern the new PGP private-key path↔resource toggle — the only change here with runtime impact (everything else is design-time placement) — so its correctness and backward-compatibility edges were the focus:

- PGP encrypt without signing with a bound-but-unused `PrivateKeyFile` resource — must **not** resolve it (resolution is gated on `SignData`); covered by new regression tests.
- PGP decrypt — private key is always required, so it resolves unconditionally.
- Persisted/reopened workflows — the path↔resource toggle restores the previously-bound side (same `PairedInputToggle` behavior as the existing key/passphrase toggles).
- Backward compatibility — `PrivateKeyFile` is a new optional `[Browsable(false)]` `InArgument<IResource>` with `[DefaultValue(null)]`; existing XAML that binds only `PrivateKeyFilePath` is unaffected.

## Implementation

Category/placement changes are applied to **both** the ViewModel `Category` and the runtime `[LocalizedCategory]` attributes (the package keeps the classic grid and unified canvas consistent); this also resolved a pre-existing key-encoding layer mismatch (runtime `Input` vs ViewModel `Options`). The new private-key resource resolution mirrors the existing `PublicKeyFile` inline `ToLocalResource().ResolveAsync().GetAwaiter().GetResult()` pattern (these are synchronous `CodeActivity` classes) and is deliberately non-throwing so an unused key can't fail a non-signing encrypt. `ActivitiesMetadata.json` needs no change — its `Properties` arrays are empty and the ViewModel drives everything. Activity reference docs under `Packaging/docs/activities/` were updated for the new key resource pairing and the IV caution.

## Caveats / Potential Issues

- The section/placement changes are design-time only (no cryptographic behavior change), but the exact visual rendering is best confirmed by opening the activities in Studio — that step can't be automated here.
- **SonarCloud "Duplication on New Code" will fail this gate and needs a maintainer waiver.** The duplicated lines are inherent to the package's parallel-class structure (Encrypt/Decrypt × Text/File, and the public/private-key symmetry): the new `PrivateKeyFile` toggle/visibility code necessarily mirrors the existing `PublicKeyFile` code in each ViewModel, and the two KeyedHash ViewModels are near-identical. The runtime duplication was already removed by extracting `PgpFileResolver.ResolveLocalPath`; the remainder is design-time scaffolding that can't be deduped without contorting the class model, and even an aggressive refactor lands ~5–6% (still above the 3% bar) while churning stable code. Coverage on new code was addressed (excluded the design-time Text ViewModels, consistent with their File siblings — now ~99%).

## How to Test

1. `dotnet build Activities/Activities.Cryptography.sln` — clean.
2. `dotnet test Activities/Activities.Cryptography.sln` — green (404 package + 123 API tests), including new reflection existence tests for `PrivateKeyFile` and two regression tests asserting a bound `PrivateKeyFile` resource is never resolved when `SignData = false`.
3. In Studio, open Encrypt/Decrypt Text & File and confirm sections render **Input → Advanced → Encoding → Options → Output**, Overwrite under Options, and the private-key "Use file / Use file path" menu action.

Independently cross-validated with Codex (2 rounds, final verdict Green — no outstanding findings).

https://claude.ai/code/session_01Wn28po9dSpgQjnBU2Jh4yJ


Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
```diff
@@ -268,14 +275,7 @@ private byte[] ExecutePgpDecrypt(CodeActivityContext context, byte[] encrypted)
                 passphraseString = new NetworkCredential(string.Empty, secure).Password;
```
viogroza, line 275: 1. Decrypt resolves the PublicKeyFile resource unconditionally, even when VerifySignature = false (medium)

DecryptFile.cs:275 / DecryptText.cs:217 (the refactored line):
var publicKeyFilePath = PgpFileResolver.ResolveLocalPath(PublicKeyFilePath.Get(context), PublicKeyFile?.Get(context));
The public key is only consumed when VerifySignature is true, yet the resource is resolved every time. This is the exact failure class the PR deliberately guards against for the encrypt private key — the new comment in EncryptFile.cs:282 says: "resolve the IResource fallback here (not unconditionally) so a bound-but-unused private key can't fail a non-signing encrypt." The same reasoning applies to the decrypt public key, but it was left unguarded.

Failure scenario: a user sets VerifySignature = true, binds PublicKeyFile to a Storage Bucket resource, then sets VerifySignature = false. The bound value persists (hiding a field doesn't clear it). At runtime the decrypt still resolves/downloads the unused resource, adding latency and — if the resource is now unreachable — failing a decrypt that needs no public key at all. The fix is to resolve it inside a if (VerifySignature) guard, mirroring the encrypt private-key pattern.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptCryptoViewModelBase.cs
```diff
@@ -54,6 +55,14 @@ protected DecryptCryptoViewModelBase(IDesignServices services) : base(services)
             {
                 AfterSwitch = ApplyPublicKeyVisibility,
             };
```
viogroza, line 57: Found by claude, I'm not sure if needs to be addressed

2. ResolveLocalPath uses IsNullOrEmpty while the sibling ResolveAsync uses IsNullOrWhiteSpace (low)

Helpers/PgpFileResolver.cs:48:
if (!string.IsNullOrEmpty(filePath) || resource == null)
    return filePath;
vs. ResolveAsync at line 25 which uses IsNullOrWhiteSpace. For the same conceptual path↔resource precedence, the synchronous activities (Encrypt/Decrypt Text & File) and the standalone async activities now disagree on whitespace.

Failure scenario: a binding yields a whitespace-only path (" ") while a valid resource is also bound. ResolveLocalPath treats " " as non-empty, returns it, never falls back to the resource, and the whitespace string is handed to file I/O → a confusing "file not found" instead of using the resource. The async path would have fallen back to the resource. Align on IsNullOrWhiteSpace.

AlexMarinescuUiPath (approved):
a few minor findings, fix if they are meaningful

Findings

Medium: missing positive test coverage for the resource-fallback branch in the new resolver logic.
The new helper resolves a local path from IResource when filePath is empty in PgpFileResolver.cs:46.
But tests only cover path-present and null-resource paths in Activities/Cryptography/UiPath.Cryptography.Activities.Tests/Helpers/PgpFileResolverTests.cs, and the added workflow tests explicitly assert non-resolution when signing is false in PgpTests.cs:227-254.
Risk: a regression in the resource conversion path could slip through without a failing test.

Low: inline API contract comment says non-throwing, but the implementation can throw from resource resolution.
Comment says non-throwing in PgpFileResolver.cs:42-44, while PgpFileResolver.cs:59-63 can propagate exceptions from ToLocalResource/ResolveAsync.
This is a documentation/maintainability mismatch rather than a runtime bug.

Activities/Cryptography/UiPath.Cryptography.Activities/DecryptFile.cs
alexandru-petre: Done in 31989a3 — `DecryptFile`/`DecryptText` now resolve the `PublicKeyFile` resource only inside `if (VerifySignature)`, mirroring the encrypt private-key guard, so a bound-but-unused public key no longer fails or slows a non-verifying decrypt. Added `PgpDecryptText/File_VerifySignatureFalse_DoesNotResolvePublicKeyResource` tests.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptCryptoViewModelBase.cs
alexandru-petre: Done in 31989a3 — `ResolveLocalPath` now short-circuits on `!IsNullOrWhiteSpace(filePath)`, matching the sibling `ResolveAsync`, so a whitespace-only path falls back to a bound resource instead of being handed to file I/O. Added a whitespace-fallback test.

alexandru-petre:
@AlexMarinescuUiPath both review-body findings addressed in 31989a3:
- **Missing positive coverage for the resource-fallback branch** — added `ResolveLocalPath_EmptyPath_WithResource_ResolvesToResourceLocalPath` (and a whitespace-fallback variant) exercising `ToLocalResource`/`ResolveAsync` via a real `LocalResource`, and dropped the now-inaccurate `[ExcludeFromCodeCoverage]` so the branch counts.
- **"Non-throwing" doc mismatch** — reworded the `ResolveLocalPath` summary: the precedence/validation is non-throwing, but resolving a supplied resource can still throw.

Note: the SonarCloud "duplication on new code" gate is intentionally left out of scope for this round.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/EncryptCryptoViewModelBase.cs
```diff
@@ -251,6 +267,11 @@ protected void ConfigureTailProperties(ref int orderIndex)
         /// </summary>
         protected void ConfigurePublicKeyFileMenuActions() => _publicKeyFileToggle.ConfigureMenuActions();
 
+        /// <summary>
+        /// Registers Main-menu actions to toggle between PrivateKeyFilePath (string) and PrivateKeyFile (IResource).
+        /// </summary>
+        protected void ConfigurePrivateKeyFileMenuActions() => _privateKeyFileToggle.ConfigureMenuActions();
```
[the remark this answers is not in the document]
alexandru-petre: Already handled by design: for these PGP-only key inputs, visibility is deliberately deferred to `AlgorithmChanged_Action` (see the doc note *"Final visibility is set by AlgorithmChanged_Action"*), which runs on designer load/restore and reads `_privateKeyFileToggle.UseSecondary` — set moments earlier by `ConfigureMenuActions`. So the restored toggle side does take effect immediately, without an inline `ApplyPrivateKeyVisibility()`. This is the same deferred pattern as `ConfigurePassphraseInputModeMenuActions`; the inline `Apply…` call is only used for the always-relevant symmetric key / input-file toggles. Resolving as already addressed.

Activities/Cryptography/UiPath.Cryptography.Activities/NetCore/ViewModels/DecryptCryptoViewModelBase.cs
```diff
@@ -229,6 +245,11 @@ protected void ConfigureTailProperties(ref int orderIndex)
         /// </summary>
         protected void ConfigurePublicKeyFileMenuActions() => _publicKeyFileToggle.ConfigureMenuActions();
 
+        /// <summary>
+        /// Registers Main-menu actions to toggle between PrivateKeyFilePath (string) and PrivateKeyFile (IResource).
+        /// </summary>
+        protected void ConfigurePrivateKeyFileMenuActions() => _privateKeyFileToggle.ConfigureMenuActions();
```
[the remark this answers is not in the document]
alexandru-petre: Already handled by design: for these PGP-only key inputs, visibility is deliberately deferred to `AlgorithmChanged_Action` (see the doc note *"Final visibility is set by AlgorithmChanged_Action"*), which runs on designer load/restore and reads `_privateKeyFileToggle.UseSecondary` — set moments earlier by `ConfigureMenuActions`. So the restored toggle side does take effect immediately, without an inline `ApplyPrivateKeyVisibility()`. This is the same deferred pattern as `ConfigurePassphraseInputModeMenuActions`; the inline `Apply…` call is only used for the always-relevant symmetric key / input-file toggles. Resolving as already addressed.

Activities/Cryptography/UiPath.Cryptography.Activities/Properties/UiPath.Cryptography.Activities.resx
```diff
@@ -913,6 +937,14 @@
...
     <comment>Category name</comment>
   </data>
+  <data name="Category_Advanced_Name" xml:space="preserve">
+    <value>Advanced</value>
+    <comment>Category name</comment>
+  </data>
+  <data name="Category_Encoding_Name" xml:space="preserve">
+    <value>Encoding</value>
+    <comment>Category name</comment>
+  </data>
```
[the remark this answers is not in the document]
alexandru-petre: That's the job of the Localization team

PR #589, opened 2026-07-23
ci: migrate Sonar analysis to @3 tasks and Java 21

alexandru-petre:
## Context
The activity packs run SonarCloud analysis through **vendored** pipeline templates under `Activities/.pipelines/templates/` (not the shared `@common` repo). The scan is split across two stages: `SonarCloudPrepare` (begin) runs in the Build stage, while `SonarCloudAnalyze`/`SonarCloudPublish` (end) run in a separate `PublishSonar` job — homemade upload/download/patch steps ferry the scanner and its state between them. All six packs share this chain via `stage.start.yml -> stage.sonar.yml`.

## Problem Statement
SonarQube Cloud no longer supports Java 17 and fails the analysis: *"The version of Java (17) ... is deprecated ... upgrade to Java 21 or later."* Every pack's Sonar stage is red.

## Behavior Before This PR
The Sonar stage runs the legacy `SonarCloud*@1` tasks with no JDK selection; the scanner picks the agent's default Java 17 and the end step aborts with the deprecation error. No coverage or quality gate is published.

## Behavior After This PR
The `@3` tasks run the analysis on Java 21 (`jdkversion: JAVA_HOME_21_X64`); the scanner and its v3 task state cross the Build -> PublishSonar boundary intact, the end step completes, and the quality gate publishes.

## Implementation
Bumping `@1 -> @3` alone was insufficient: `SonarCloudPrepare@3` moved its cross-task variables to the `SONAR_*` namespace (`SONAR_SCANNER_MODE`, `SONAR_ENDPOINT`, `SONAR_SCANNER_REPORTTASKFILE`, `SONAR_SERVER_VERSION`; only `SONARQUBE_SCANNER_PARAMS` kept its old name) and its scanner-path vars (`SONAR_SCANNER_DOTNET_EXE/DLL`). Because this repo splits the tasks across stages, the upload/download/transfer steps were updated to carry the v3 names — `Analyze@3` hard-fails without `SONAR_SCANNER_MODE`, `Publish@3` without `SONAR_ENDPOINT`. Names verified against SonarSource's `sonarcloud-v3` `TaskVariables` enum. The report-task-file is re-pointed to the PublishSonar job's temp dir **without** `isOutput`, so the same-job tasks read the rewrite.

## Caveats / Potential Issues
- **Extension prerequisite:** the `@3` tasks require the *SonarQube Cloud* (v3) AzDO extension installed org-wide, or the pipeline fails to compile (*"A task is missing"*).
- **`PatchSonarQubeTask@0.7921.14273`** pin (stage.sonar.yml) is unchanged — confirm it stays valid against the installed extension payload.
- Cross-stage propagation of the secret `SONAR_ENDPOINT` could not be validated statically; it needs a real run.

## How to Test

- [x] Trigger a pack pipeline (push a change under `Activities/Java/*`, or queue it manually) and confirm: (1) the pipeline **compiles**, and (2) the `RunSonar` job's "Run Code Analysis" step completes without the Java-version error and the quality gate publishes.

### DIFF
--- a/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
@@ -3,7 +3,7 @@ parameters:
   jdkVersion: JAVA_HOME_21_X64
 
 steps:
-- task: SonarCloudAnalyze@3
+- task: SonarCloudAnalyze@3.4.3
   displayName: 'Run Code Analysis'
   condition: ${{ parameters.condition }}
   inputs:
--- a/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
@@ -1,7 +1,10 @@
 parameters:
   condition: succeeded()
+  jdkVersion: JAVA_HOME_21_X64
 
 steps:
-- task: SonarSource.sonarcloud.ce096e50-6155-4de8-8800-4221aaeed4a1.SonarCloudAnalyze@1
+- task: SonarCloudAnalyze@3
   displayName: 'Run Code Analysis'
   condition: ${{ parameters.condition }}
+  inputs:
+    jdkversion: ${{ parameters.jdkVersion }}
--- a/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
@@ -1,9 +1,10 @@
 parameters:
   condition: succeeded()
   jdkVersion: JAVA_HOME_21_X64
+  sonarTaskVersion: ''
 
 steps:
-- task: SonarCloudAnalyze@3.4.3
+- task: SonarCloudAnalyze@${{ parameters.sonarTaskVersion }}
   displayName: 'Run Code Analysis'
   condition: ${{ parameters.condition }}
   inputs:
--- a/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
@@ -1,6 +1,6 @@
 parameters:
   condition: succeeded()
-  jdkVersion: JAVA_HOME_21_X64
+  jdkVersion: ''
   sonarTaskVersion: ''
 
 steps:
--- a/Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
@@ -21,7 +21,7 @@ steps:
       condition: ${{ parameters.condition }}
 
 - ${{ if ne(parameters.configFile, '') }}:
-  - task: SonarCloudPrepare@3
+  - task: SonarCloudPrepare@3.4.3
     displayName: 'Prepare analysis on SonarCloud'
     condition: ${{ parameters.condition }}
     inputs:
@@ -38,7 +38,7 @@ steps:
       coverageExclusions: ${{ parameters.coverageExclusions }}
       runsettingsFile: ${{ parameters.runsettingsFile }}
 
-  - task: SonarCloudPrepare@3
+  - task: SonarCloudPrepare@3.4.3
     displayName: 'Prepare analysis on SonarCloud'
     condition: ${{ parameters.condition }}
     inputs:
--- a/Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
@@ -21,7 +21,7 @@ steps:
       condition: ${{ parameters.condition }}
 
 - ${{ if ne(parameters.configFile, '') }}:
-  - task: SonarSource.sonarcloud.14d9cde6-c1da-4d55-aa01-2965cd301255.SonarCloudPrepare@1
+  - task: SonarCloudPrepare@3
     displayName: 'Prepare analysis on SonarCloud'
     condition: ${{ parameters.condition }}
     inputs:
@@ -38,7 +38,7 @@ steps:
       coverageExclusions: ${{ parameters.coverageExclusions }}
       runsettingsFile: ${{ parameters.runsettingsFile }}
 
-  - task: SonarSource.sonarcloud.14d9cde6-c1da-4d55-aa01-2965cd301255.SonarCloudPrepare@1
+  - task: SonarCloudPrepare@3
     displayName: 'Prepare analysis on SonarCloud'
     condition: ${{ parameters.condition }}
     inputs:
--- a/Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
+++ b/Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
@@ -13,6 +13,7 @@ parameters:
   logLevel: 'Info'
   verbose: 'false'
   cleanup: true
+  sonarTaskVersion: ''
 
 steps:
 - ${{ if eq(parameters.cleanup, true) }}:
@@ -21,7 +22,7 @@ steps:
       condition: ${{ parameters.condition }}
 
 - ${{ if ne(parameters.configFile, '') }}:
-  - task: SonarCloudPrepare@3.4.3
+  - task: SonarCloudPrepare@${{ parameters.sonarTaskVersion }}
     displayName: 'Prepare analysis on SonarCloud'
     condition: ${{ parameters.condition }}
     inputs:
@@ -38,7 +39,7 @@ steps:
       coverageExclusions: ${{ parameters.coverageExclusions }}
       runsettingsFile: ${{ parameters.runsettingsFile }}
 
-  - task: SonarCloudPrepare@3.4.3
+  - task: SonarCloudPrepare@${{ parameters.sonarTaskVersion }}
     displayName: 'Prepare analysis on SonarCloud'
     condition: ${{ parameters.condition }}
     inputs:
--- a/Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
+++ b/Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
@@ -39,15 +39,30 @@ steps:
         Write-Error "The SonarQube directory does not exist!"
       }
 
+      # SonarCloudPrepare@3 publishes v3 scanner-path variable names
+      # (SONAR_SCANNER_DOTNET_*), while @1 published SONARQUBE_SCANNER_MSBUILD_*.
+      # Probe both families so the scanner staging works across task versions.
+      $sonarEnvVars = @(
+        'SONARQUBE_SCANNER_DOTNET_DLL',
+        'SONARQUBE_SCANNER_DOTNET_EXE',
+        'SONARQUBE_SCANNER_MSBUILD_DLL',
+        'SONARQUBE_SCANNER_MSBUILD_EXE',
+        'SONAR_SCANNER_DOTNET_DLL',
+        'SONAR_SCANNER_DOTNET_EXE',
+        'SONAR_SCANNER_MSBUILD_DLL',
+        'SONAR_SCANNER_MSBUILD_EXE'
+      )
       $sonarScannerDir = ""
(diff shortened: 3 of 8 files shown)


Activities/.pipelines/templates/stage.sonar.yml
```diff
@@ -45,6 +49,34 @@ jobs:
         repoPath: $(Build.SourcesDirectory)
         scannerPath: "$(Build.SourcesDirectory)\\SonarTemp\\scanner"
 
+    # SonarCloudPrepare@3 records the report-task.txt path against the Build job's
+    # Agent.TempDirectory; rewrite it to this (PublishSonar) job's temp path so
+    # SonarCloudPublish@3 can locate the report and poll the Quality Gate. No
+    # isOutput here: the value must overwrite the in-job SONAR_SCANNER_REPORTTASKFILE
+    # env that the analyze/publish tasks read later in this same job.
+    - powershell: |
+        $patchedReportPath = "$env:SONAR_SCANNER_REPORTTASKFILE".Replace("$(OriginalTempPath)", "$(Agent.TempDirectory)")
```
LiviuPonova: **Finding: missing `RunAnalysis` condition guard + potential `ArgumentException` on empty path**

Every other Sonar-related step in this file is gated behind `condition: eq(variables['RunAnalysis'], 'true')`, but this step isn't. It also does `"$env:SONAR_SCANNER_REPORTTASKFILE".Replace("$(OriginalTempPath)", ...)` — if `OriginalTempPath` ever resolves to an empty string (e.g. an upstream step that's supposed to set it didn't run/emit output), `.Replace("", ...)` throws `ArgumentException` and fails the "Analyze & publish results" job outright, even in scenarios where analysis wasn't meant to run at all.

Suggest adding the same `RunAnalysis` condition as the sibling steps, and/or guarding against an empty `OriginalTempPath` before the `.Replace()` call.

Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
```diff
@@ -21,7 +21,7 @@ steps:
       condition: ${{ parameters.condition }}
 
 - ${{ if ne(parameters.configFile, '') }}:
-  - task: SonarSource.sonarcloud.14d9cde6-c1da-4d55-aa01-2965cd301255.SonarCloudPrepare@1
+  - task: SonarCloudPrepare@3.4.3
```
LiviuPonova: **Finding: SonarCloud task version `3.4.3` hardcoded 4x across 3 files, no single source of truth**

The exact task version `3.4.3` is repeated literally in 4 places: here and at line 41 of this file, plus `SonarCloudAnalyze@3.4.3` in `analyze-sonar-coverage.yml:6` and `SonarCloudPublish@3.4.3` in `publish-sonar-coverage.yml:7`.

`stage.sonar.yml` (a few lines above the `PatchSonarQubeTask@0.7921.14273` step) has a comment documenting a prior outage from exactly this failure mode: a pinned task version got out of sync and the pipeline broke with "A task is missing" (commit b587daa). The same risk now applies to `3.4.3` — a future extension bump means hand-editing all 4 spots, and missing one reproduces that outage.

Consider hoisting the version into a single parameter/variable shared by all three templates.

Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
```diff
@@ -39,17 +39,32 @@ steps:
         Write-Error "The SonarQube directory does not exist!"
       }
 
+      # SonarCloudPrepare@3 publishes v3 scanner-path variable names
+      # (SONAR_SCANNER_DOTNET_*), while @1 published SONARQUBE_SCANNER_MSBUILD_*.
+      # Probe both families so the scanner staging works across task versions.
+      $sonarEnvVars = @(
```
LiviuPonova: **Finding: scanner env-var alias list duplicated across two files**

This 8-name list of possible scanner-path variable names has to stay in lockstep with `download-sonar-build-output.yml`, which sets these same 8 names (4 per branch, DLL vs EXE) a few lines up in that file. They're hand-synced across two separate files rather than sharing one source of truth.

If a naming convention is ever added or dropped in only one of the two files, the lookup here can silently fail to find a variable that really was set — which trips the new `throw` at line 65 even though the scanner was found successfully by the producing step.

Consider extracting the alias list into a single shared template/variable both files reference.

Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
```diff
@@ -1,7 +1,10 @@
 parameters:
   condition: succeeded()
+  jdkVersion: JAVA_HOME_21_X64
```
LiviuPonova: **Finding: JDK version hardcoded in two unlinked places**

`JAVA_HOME_21_X64` is hardcoded here as the parameter default, and again as a literal `$env:JAVA_HOME_21_X64` inside the `JavaExePath` patch step in `stage.sonar.yml` (line 70) — the very step this PR added to fix the agent/JDK mismatch.

The two aren't wired together. If the JDK version is bumped later by changing this default, the hardcoded reference in `stage.sonar.yml` won't move with it, so `SonarCloudAnalyze` would run under the new JDK while `JavaExePath` still gets patched to point at the old one — reintroducing the exact mismatch this PR fixes.

Worth passing the JDK var name through as a pipeline variable/parameter that both spots read, rather than hardcoding it twice.

LiviuPonova:
**Non-blocking architectural observation: repeated ad hoc path-patching across split Prepare/Analyze/Publish jobs**

Not a defect in this PR — just a pattern worth flagging for the team to weigh, since it keeps recurring.

This pipeline splits Sonar work across separate jobs/agents: "Prepare" runs on one agent (downloading the scanner, generating config files with that agent's local paths baked in), while "Analyze" and "Publish" run later on a different agent. Whenever the generated config references an agent-specific path, it breaks across that agent boundary, and the fix so far has been a bespoke find/replace "patch" step per broken path.

This PR adds two more such patches in `stage.sonar.yml` (the `SONAR_SCANNER_REPORTTASKFILE` path around line 57, and the `JavaExePath` patch around line 68), on top of the pre-existing `PatchSonarQubeTask` step. That's 3 separate hand-written patches now, each targeting a different property that broke for the same underlying reason.

Functionally this is fine and the PR's fixes look reasonable on their own. Just flagging that every future SonarScanner/SonarCloud extension change that bakes in a new agent-local path will likely need patch #4, #5, etc. It might be worth a future discussion on whether running prepare/analyze/publish in the same job/agent (or another way of avoiding the cross-agent path issue at the source) would be more sustainable than continuing to add patches — but that's an architectural call for the team, not something this PR needs to solve.

Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
```diff
@@ -39,17 +39,32 @@ steps:
...
+      foreach ($var in $sonarEnvVars) {
+        $value = [System.Environment]::GetEnvironmentVariable($var)
+        if (![string]::IsNullOrEmpty($value)) {
+          $sonarScannerDir = [System.IO.Path]::GetDirectoryName($value)
+          break
         }
       }
 
+      if ([string]::IsNullOrEmpty($sonarScannerDir)) {
+        throw "Could not find the Sonar Scanner directory - none of the expected scanner-path variables were set by SonarCloudPrepare. Aborting before the scanner copy."
```
LiviuPonova, line 58: **Finding: fail-fast `throw` is neutered by `continueOnError: true`**

This new `throw` for a missing Sonar Scanner directory sits inside a task that has `continueOnError: true` set (line 72, pre-existing). With `continueOnError: true`, Azure Pipelines marks the step "succeeded with issues" and continues the job even when the script throws — so this hardening can never actually fail the pipeline. If the scanner path lookup fails, the job proceeds, publishes an incomplete Sonar artifact, and the real problem only surfaces later in a downstream job (RunSonar) with a more confusing error.

Suggest either removing `continueOnError: true` from this task, or moving this validation into a separate step that isn't marked continue-on-error.

LiviuPonova (approved):
Approving — the overall direction (fixing the JDK-21/agent mismatch and hardening the scanner-path lookup) is solid, and I left 6 review comments (5 inline + 1 general) with suggested follow-ups.

One item is worth calling out explicitly before merge: the new fail-fast `throw` in `upload-sonar-build-output.yml` (missing Sonar Scanner directory) currently sits inside a step marked `continueOnError: true`, so it can never actually fail the job — the intended hardening doesn't take effect yet. Please take a look at that one in particular; the rest are lower-severity cleanup/maintainability notes (hardcoded task version repeated across files, a duplicated env-var alias list, JDK version hardcoded in two places) plus one non-blocking architectural observation about the growing number of ad hoc path patches across the split Prepare/Analyze/Publish jobs.

Approving on the basis that these get addressed before/at merge.

Activities/.pipelines/templates/stage.sonar.yml
alexandru-petre: Fixed in 36ad237: added the RunAnalysis condition to match the sibling steps, and guarded the Replace() call against an empty OriginalTempPath (it now skips the rewrite and logs, instead of throwing ArgumentException).

Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
alexandru-petre: Addressed in 36ad237: hoisted the version to a single sonarTaskVersion parameter in stage.start.yml (default 3.4.3), threaded down to prepare/analyze/publish so there is only one place to bump on an extension update. Leaving this thread open until a pipeline run confirms the template-parameter reference compiles in the task-version position.

alexandru-petre:
Fixed in 36ad237: added a separate **Verify Sonar scanner was staged** step (without `continueOnError`) right after the copy step, so a missing scanner now fails the job instead of publishing an incomplete artifact. I left the copy step's pre-existing `continueOnError: true` untouched to avoid changing the behavior of its other branches (RoslynCA copy / symlink).

Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
alexandru-petre: Addressed in 36ad237 by a separate step rather than touching this one: I added a Verify Sonar scanner was staged step immediately after this task, without continueOnError, so a missing scanner now fails the job there. I intentionally left this task pre-existing continueOnError untouched because it also guards the RoslynCA-copy and symlink branches. This throw still serves as an early-exit that avoids the empty Copy-Item within this step. Happy to drop the now-redundant throw if you prefer.

Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
alexandru-petre: Addressed in a95b235: hoisted the scanner-path variable names to sonarScannerVarsDll / sonarScannerVarsExe in stage.start.yml (single source), threaded to both the upload probe and the download republish, which now derive their lists from those params via Split. Leaving this open until a pipeline run confirms the parameterization compiles.

Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
alexandru-petre: Addressed in a95b235: hoisted the JDK env-var name to sonarJdkVersion in stage.start.yml, now used both by the analyze jdkVersion input and by the JavaExePath patch step in stage.sonar.yml (one place to bump). Leaving this open until a pipeline run confirms it compiles.

Activities/.pipelines/templates/Sonar/prepare-sonar-coverage.yml
alexandru-petre: Validated: build 12795849 (a95b235) compiled and the full Sonar stage passed, confirming @${{ parameters.sonarTaskVersion }} is honored in the task-version position. Resolving.

Activities/.pipelines/templates/Sonar/upload-sonar-build-output.yml
alexandru-petre: Validated: build 12795849 passed with the scanner-var lists sourced from stage.start.yml and Split in both upload and download (Copy Sonar files + Set SonarQube Scanner Path both green). Resolving.

Activities/.pipelines/templates/Sonar/analyze-sonar-coverage.yml
alexandru-petre: Validated: build 12795849 passed with sonarJdkVersion driving both the analyze jdkVersion input and the JavaExePath patch step (both green). Resolving.

PR #590, opened 2026-08-04
FTP: unified-canvas UX, legacy designer removal, SFTP enumeration fixes [STUD-80716]

viogroza:
Adopts the unified canvas across all 8 FTP activities, removes the legacy WPF designer project, and fixes the reported SFTP enumeration crash. Follows the Cryptography pack (`c21a272`, STUD-80718) as the reference implementation.

## Unified canvas

- **Property configuration moved out of `ActivitiesMetadata.json` into the ViewModels.** The JSON now declares activities only (`"Properties": []`), matching Cryptography. `WithFtpSession`'s `Body` is declared as a `DesignProperty` and hidden in the ViewModel, the same way `JavaScopeViewModel` handles its scope body.
- **One section vocabulary across both layers:** Input -> Options -> Security / Proxy (scope only) -> Output last, applied to the ViewModels *and* the runtime `[LocalizedCategory]` attributes so the two cannot drift.
- Toggle for every boolean (including `ContinueOnError`), Number for `Port` / `Timeout` / `ProxyPort`, a data source for `FtpsMode`, placeholders on path fields.
- **Verb-first renames:** Directory Exists -> **Check If Folder Exists**, File Exists -> **Check If File Exists**, Enumerate Objects -> **List Files and Folders**. Old names keep working as search aliases via `DisplayNameAliasKeys`.
- Three field-anchored design-time validations on `WithFtpSession` (missing username, no authentication method, negative timeout).

## Legacy designer removal

`UiPath.FTP.Activities.Design` deleted (40 files) - activities render from their ViewModels now, mirroring Cryptography's `feb1052`. Packaging narrows to `$(PortableFramework)`; the Windows TFM existed only to carry the Design dll. Activity search synonyms were preserved through the metadata JSON.

## SFTP enumeration fix

The reported failure - `SftpPathNotFoundException: No such file` when enumerating `"."`:

- **An unreadable sub-directory no longer discards the whole enumeration.** Servers can return an entry from `readdir` and then refuse `opendir` on it. That sub-directory is now skipped and its path traced, instead of failing the operation. Connection and timeout failures still propagate, and the caller's own `RemotePath` is not covered by the leniency.
- **Dropped the `ChangeDirectory` / restore dance.** It was not in a `finally`, so a mid-walk failure left the shared session's working directory pointing into the tree - and since the scope shares one session, with `ContinueOnError = True` every later relative path resolved from the wrong place.
- 16 listing tests added (11 enumeration, 5 download).

## Behaviour changes reviewers should look at

1. **The enumeration leniency above** - by CONTRIBUTING's wording ("an exception is no longer thrown") this is a Bucket 1 change. It is the point of the fix, but calling it out explicitly. Affects **List Files and Folders** and **Download Files** on SFTP only; Delete, Upload, Move and the Exists pair are untouched, as is plain FTP/FTPS.
2. **Download Files can now report success on a partial transfer** when a sub-folder is skipped. The trace names the path; the activity outcome does not. Documented.
3. **The new design-time validations change exception type and timing for an already-broken scope** - `InvalidWorkflowException` at workflow load instead of `ArgumentNullException` during execution, which an in-workflow Try/Catch can no longer intercept. No previously-*working* workflow is affected: each validation fires only where `ExecuteAsync` already threw.

**Symbolic-link handling is unchanged** across all 8 activities - the recursion and file-selection predicates are equivalent to the pre-branch ones. **No public API change:** no activity property added, removed, renamed or retyped; no `[RequiredArgument]` / `[Obsolete]` / `[DefaultValue]` change; no design-time `IsRequired` change; `UiPath.FTP.Activities.API` untouched.

## Known limitation - documented, pinned, deliberately not fixed here

`FtpObjectType.Link` is unreachable on the SFTP path. SSH.NET's `IsRegularFile` is a `(mode & S_IFREG) == S_IFREG` test and `S_IFLNK` contains `S_IFREG`, so every symbolic link classifies as `File`: **List Files and Folders' `Link` filter matches nothing over SFTP**, and `Check If File Exists` answers `True` for a link while `Check If Folder Exists` answers `False` for a linked folder. FTP/FTPS report `Link` correctly, so the two protocols disagree.

The fix is understood - separate the dispatch classifier from the reporting one; a naive reorder regresses `Download` and `FileExists` - but it changes `FtpObjectInfo.Type` for shipped workflows, so it belongs in its own change with its own release note. The shipped docs now state it everywhere they previously promised `Link`, and `SFTP_GetFtpObjectType_ClassifiesSymbolicLinkAsFile` pins it so it cannot be changed by accident.

## Verification

- `Activities.FTP.sln` builds clean - 0 errors; the only warnings are 2 pre-existing `CS0618` (`FtpSslProtocols.Default`, x2 TFMs) and 2 NuGet pack deprecations.
- `dotnet test Activities.FTP.sln`: **59/59** in `UiPath.FTP.Tests`, 18/19 in `UiPath.FTP.Activities.API.Tests` (1 pre-existing skip).
- Package verified: single `lib/net6.0` with `UiPath.FTP.dll`, `UiPath.FTP.Activities.dll`, `.API.dll/.xml` and 13 satellite locales - no Design dll, no `lib/net6.0-windows`. Same layout Cryptography has shipped since `feb1052`.
- Every metadata / `.resx` / C# resource key resolves; all 8 icons present; runtime `[LocalizedCategory]` matches ViewModel `Category` on every visible property.

## Before merge

- [x] **Not yet opened in Studio.** This is a UX change and it has only been verified structurally. A canvas smoke test (and screenshots) is the main gap.
- [ ] **Localization sync required.** 52 changed strings still carry their *old* translations in all 13 locales, and 18 new keys have none. Missing keys fall back to English; stale ones do not - a German user still sees "Directory Exists". Also worth excluding the deleted `UiPath.FTP.Activities.Design.*.resx` path on the Localization side so the orphans do not come back.
- [ ] **Version bump?** Release `VersionPrefix` is still `3.2.0`. The package now ships one fewer assembly and drops a TFM. Neither reference commit touched the version props, so precedent says leave it - flagging for a maintainer call.

## Follow-up, needs a repo-owner decision

`UseSftp` (bool) + `FtpsMode` (3-value enum) are two controls for one concept: the transport is really a four-way choice, so 6 states encode 4 meanings and `FtpsMode` is silently ignored when `UseSftp` is on. Consolidating into one `Protocol` selector touches public properties, so it needs owner sign-off. A cheap non-breaking slice exists in the meantime: there is no rule on `UseSftp`, so with SFTP selected the user still sees `FtpsMode`, `SslProtocols` and `AcceptAllCertificates`, none of which SFTP reads.


Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
```diff
@@ -184,6 +185,28 @@ protected override void CacheMetadata(NativeActivityMetadata metadata)
                 metadata.AddValidationError(new ValidationError(string.Format(Resources.ValidationErrorFormat, Resources.Activity_WithFtpSession_Property_ProxyServer_Name), false, nameof(ProxyServer)));
             if (ProxyType != FtpProxyType.None && ProxyPort?.Expression == null)
                 metadata.AddValidationError(new ValidationError(string.Format(Resources.ValidationErrorFormat, Resources.Activity_WithFtpSession_Property_ProxyPort_Name), false, nameof(ProxyPort)));
+
+            // Surface at design time what ExecuteAsync would otherwise only throw at runtime, so the
+            // fields get an error indicator and an entry in the Issues panel.
+            if (!UseAnonymousLogin)
```
AlexMarinescuUiPath: `UseAnonymousLogin` cannot be combined with `UseSftp`, but this branch skips every authentication validation for that combination. `ExecuteAsync` also leaves `ftpConfiguration.Username` unset when anonymous login is enabled, while `SftpSession` requires a username plus password/keyboard-interactive or private-key authentication and throws `NoValidAuthenticationMethod` when no method exists. As a result, anonymous SFTP passes `CacheMetadata` and always fails while opening the session. Please add an explicit validation error for `UseSftp && UseAnonymousLogin` and cover it with a validation test.

AlexMarinescuUiPath (changes requested):
Requesting changes for the following issues:

1. Copilot already identified that the conditional visibility handlers are registered but never applied during `InitializeModel` (`Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs:158-189`, existing comment: https://github.com/UiPath/Community.Activities/pull/590#discussion_r3720831015). Restored/default password, FTPS, and proxy modes can therefore render the wrong fields until a controlling value changes.

2. The new validation at `Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs:188-209` does not reject anonymous SFTP. The inline comment explains the runtime failure path.

3. The unified-canvas migration adds reactive rule and dependency wiring at `Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs:366-385`, but there are no FTP ViewModel tests for initial state, restored values, or mode transitions. This gap already allowed issue 1 through, and Sonar reports only 12.5% coverage on new code against the 80% gate. Please add focused ViewModel and validation tests with the fixes.

The runtime FTP suite passes locally (154 passed, 2 real-server tests skipped), so these requests are focused on design-time correctness and coverage.

Activities/FTP/UiPath.FTP.Activities/NetCore/ViewModels/WithFtpSessionViewModel.cs
```diff
@@ -188,6 +188,180 @@ protected override void InitializeModel()
               .BuildAndInsertMenuActions(true);
         }
```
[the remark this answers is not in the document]
viogroza: will do, ty

Activities/FTP/UiPath.FTP.Activities/WithFtpSession.cs
viogroza: will do, ty

AlexMarinescuUiPath (approved):
Re-reviewed commits `eff7128` and `0c039be` against the previously reviewed head.

The requested changes are addressed:
- Copilot's initial-visibility finding is now covered by real `InitializeAsync()` tests for default and restored password, FTPS, and proxy states.
- Anonymous SFTP is rejected in `CacheMetadata`, including when other credentials are present, with focused validation tests.
- All eight migrated ViewModels now have targeted configuration coverage, and the Sonar quality gate is green.

I also checked the Sonar-driven `SftpSession` and `WithFtpSession` refactors for authentication, exception, cancellation, overwrite, scheduling, and disposal drift; no new blocking issue found. Local result: 230 passed, 2 real-server tests skipped, 0 failed across both target frameworks.

Residual UX risk: the PR still notes that it has not been smoke-tested in Studio. The ViewModel tests validate state and rule wiring, but not host renderer/layout parity, so that manual check should still happen before merge.

viogroza:
basic testing was done in studio

PR #596, opened 2026-08-19
Python: Fix venv isolation from user-site and base-install site-packages [STUD-81085]

viogroza:
## Summary
- Python Scope's embedded interpreter never goes through CPython's own venv activation path (`site.py`'s `venv()` function only triggers for a normally-launched `<venv>/Scripts/python.exe`), so pointing `Path` at a venv left both the PEP-370 user-site directory and the base install's own site-packages on `sys.path` unfiltered.
- A native package (e.g. pywin32) installed with a different build in either location could resolve its Python module from one and its native DLL dependency from the other, mismatched, copy — surfacing as `DLL load failed while importing win32api: The specified procedure could not be found.`
- Fixes both leak paths via `PythonEngine.SetNoSiteFlag()` (gated on whether the venv requests `--system-site-packages`). For a `--system-site-packages` venv, an ambient `PYTHONNOUSERSITE` in the host's own environment is now honored exactly as a natively-activated interpreter would (no override).
- Also fixes `PythonHome` being pointed at the venv root instead of its declared base install (breaks stdlib resolution entirely) — falling back to `LibraryPath`'s own directory when the declared base install doesn't actually carry a stdlib (e.g. a Microsoft Store Python) — tightens venv detection against false positives and a trailing-separator miss, and adds a version cross-check between a venv and the Python library actually being loaded.
- Fixes `sitecustomize.py`/`usercustomize.py` precedence for `--system-site-packages` venvs so the venv's own copy gets a chance to run, without re-running a base/user copy's side effects a second time when the venv has none of its own.

## Behaviour changes
- `EngineProvider.ValidateInstallation` (called from `PythonScope.ExecuteAsync`/`PythonService`) now throws `NotSupportedException` when a venv's declared Python version (`pyvenv.cfg`) doesn't match the version actually loaded from `LibraryPath`. Previously this pairing would proceed and fail later with a confusing, version-specific stdlib error.

## Test plan
- [x] `UiPath.Python.Tests`, `UiPath.Python.Activities.Tests`, `UiPath.Python.Activities.API.Tests` all pass (98 total)
- [x] Regression coverage added for: user-site isolation (default + `--system-site-packages`), base-install site-packages leak, `sitecustomize.py`/`usercustomize.py` preservation and precedence (including the no-double-execution case), `PythonHome` resolution and its Microsoft-Store/no-stdlib fallback, venv detection (root/`Scripts`/`bin`/trailing-separator/false-positive/unreadable-file), venv/library version mismatch (including `version_info`-only configs), ambient `PYTHONNOUSERSITE` handling, and the Windows `Lib` vs POSIX `lib` case-sensitive path segments
- [x] Each new/changed behavior verified to actually catch its regression by temporarily reverting the fix and confirming the test fails, then restoring it

### DIFF
--- a/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
+++ b/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
@@ -283,6 +283,44 @@ namespace UiPath.Python.Tests
 
         [Fact]
         [Trait(TestCategories.Category, Category)]
+        public async Task Venv_With_SystemSitePackages_And_No_Own_SiteCustomize_Does_Not_Rerun_UserSite_Copy()
+        {
+            // Regression test for a real bug in an earlier version of this fix: unconditionally
+            // popping "sitecustomize" from sys.modules and re-importing, regardless of whether the
+            // venv itself has its own copy, would still find the *same* user-site copy again when
+            // the venv has none of its own — re-running its side effects a second time. Gating the
+            // pop+reimport on the venv's own site-packages actually containing a sitecustomize.py
+            // (see PostInitializationVenvSetup) avoids this: with no venv-local copy, the module
+            // site.main() already cached is left untouched, so it runs exactly once.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages"));
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"),
+                $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
+
+            var markerPath = _markerFile.Replace("\\", "/");
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, EmbeddedPythonRuntimeBootstrap.UserSiteVersionFolder, "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(userSitePackages, "sitecustomize.py"),
+                $"import codecs{Environment.NewLine}codecs.open('{markerPath}', 'a', encoding='utf-8').write('usersite-sitecustomize\\n'){Environment.NewLine}");
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
+            var single = Assert.Single(markerLines);
+            Assert.Equal("usersite-sitecustomize", single);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
         public async Task Venv_With_SystemSitePackages_Honors_Ambient_PYTHONNOUSERSITE()
         {
             // A --system-site-packages venv's own flag only says "also add the base install's
--- a/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
+++ b/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
@@ -43,6 +43,15 @@ namespace UiPath.Python.Tests
 
             _previousUserBase = Environment.GetEnvironmentVariable("PYTHONUSERBASE");
             _previousNoUserSite = Environment.GetEnvironmentVariable("PYTHONNOUSERSITE");
+
+            // Start every test from a known-clean baseline regardless of whatever the host
+            // machine's own ambient environment happens to have for these two — otherwise a
+            // machine/CI agent with e.g. PYTHONNOUSERSITE already set would make tests that expect
+            // user-site processing fail for reasons unrelated to the behavior under test. The one
+            // test that specifically wants an ambient value sets it itself, after this. Dispose
+            // restores whatever was actually ambient when this instance was constructed.
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", null);
+            Environment.SetEnvironmentVariable("PYTHONNOUSERSITE", null);
         }
 
         public void Dispose()
@@ -168,6 +177,41 @@ namespace UiPath.Python.Tests
 
         [Fact]
         [Trait(TestCategories.Category, Category)]
+        public async Task Venv_ExitAndHelp_Builtins_Still_Resolve()
+        {
+            // Regression test: SetNoSiteFlag skips site.main() entirely for a default venv, which
+            // does more than path setup — it also calls setquit()/setcopyright()/sethelper(),
+            // installing the exit/quit/help/copyright/credits/license builtins. Without restoring
+            // them, a script calling exit() (a common, if discouraged, RPA pattern) would die with
+            // a NameError that gives no hint it's related to venv handling — it works today, works
+            // natively in a venv, and works in every non-venv scope.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages"));
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {EmbeddedRuntimePath}{Environment.NewLine}");
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            string result;
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+
+                var script = await engine.LoadScript(
+                    "import builtins\ndef check():\n    return f\"{hasattr(builtins, 'exit')},{hasattr(builtins, 'help')}\"\n",
+                    CancellationToken.None);
+                var invoked = await engine.InvokeMethod(script, "check", null, CancellationToken.None);
+                result = (string)engine.Convert(invoked, typeof(string));
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            Assert.Equal("True,True", result);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
         public async Task Venv_Does_Not_Leak_BaseInstall_SitePackages()
         {
             // Regression test for a second, distinct leak path found by inspecting a real venv's
@@ -238,6 +282,40 @@ namespace UiPath.Python.Tests
 
             var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
             Assert.Contains("sitecustomize", markerLines);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_SiteCustomize_Still_Runs_From_BaseInstall_When_Venv_Has_None()
+        {
+            // Regression test: for a default venv (SetNoSiteFlag set) with no sitecustomize.py of
+            // its own, the base install's Lib\sitecustomize.py (e.g. corporate proxy/logging setup)
+            // must still be picked up — the stdlib paths remain on sys.path under Py_NoSiteFlag, and
+            // this must be an unconditional call for the default-venv case, unlike the
+            // --system-site-packages case where it's gated on the venv having its own copy.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            var fakeBase = Directory.CreateDirectory(Path.Combine(_rootDir, "fakebase")).FullName;
+            Directory.CreateDirectory(Path.Combine(fakeBase, "Lib", "encodings"));
+            var markerPath = _markerFile.Replace("\\", "/");
+            File.WriteAllText(Path.Combine(fakeBase, "Lib", "sitecustomize.py"),
+                $"import codecs{Environment.NewLine}codecs.open('{markerPath}', 'a', encoding='utf-8').write('base-sitecustomize\\n'){Environment.NewLine}");
+
+            Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages"));
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {fakeBase}{Environment.NewLine}");
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
+            Assert.Contains("base-sitecustomize", markerLines);
         }
 
         [Fact]
--- a/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
+++ b/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
@@ -18,18 +18,14 @@ namespace UiPath.Python.Tests
     //
     // This test doesn't need real pywin32 binaries: the bug is that the user-site .pth runs at
     // all when a venv is configured, so a .pth marker is a faithful, hermetic reproduction of the
-    // exact mechanism pywin32 relies on. Runs out-of-process (the real production default) since
-    // the fix depends on an environment variable set in the parent process before the host spawns.
+    // exact mechanism pywin32 relies on. Runs out-of-process since that's the real production
+    // default path (PythonScope.Isolated defaults to true).
     public class VenvUserSiteIsolationTests : IDisposable
     {
         private const string Category = "Python";
 
         private static readonly string EmbeddedRuntimePath = EmbeddedPythonRuntimeBootstrap.EnsureRuntimePath();
         private static readonly string EmbeddedLibraryPath = EmbeddedPythonRuntimeBootstrap.GetPythonLibraryPath(EmbeddedRuntimePath);
-
-        // Must match the running embeddable interpreter's sys.version_info (3.14.5) — Windows
-        // user-site resolves to <PYTHONUSERBASE>\Python<major><minor>\site-packages.
-        private const string UserSiteVersionFolder = "Python314";
 
         private readonly string _rootDir;
         private readonly string _venvDir;
@@ -54,6 +50,7 @@ namespace UiPath.Python.Tests
             Environment.SetEnvironmentVariable("PYTHONUSERBASE", _previousUserBase);
             Environment.SetEnvironmentVariable("PYTHONNOUSERSITE", _previousNoUserSite);
             try { Directory.Delete(_rootDir, true); } catch { /* best effort cleanup */ }
+            GC.SuppressFinalize(this);
         }
 
         [Fact]
@@ -66,7 +63,7 @@ namespace UiPath.Python.Tests
             File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {EmbeddedRuntimePath}{Environment.NewLine}");
             WriteMarkerPth(venvSitePackages, "venv");
 
-            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, UserSiteVersionFolder, "site-packages")).FullName;
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, EmbeddedPythonRuntimeBootstrap.UserSiteVersionFolder, "site-packages")).FullName;
             WriteMarkerPth(userSitePackages, "usersite");
             Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
 
@@ -108,7 +105,7 @@ namespace UiPath.Python.Tests
                 $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
             WriteMarkerPth(venvSitePackages, "venv");
 
-            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, UserSiteVersionFolder, "site-packages")).FullName;
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, EmbeddedPythonRuntimeBootstrap.UserSiteVersionFolder, "site-packages")).FullName;
             WriteMarkerPth(userSitePackages, "usersite");
             Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
 
@@ -245,16 +242,57 @@ namespace UiPath.Python.Tests
 
         [Fact]
         [Trait(TestCategories.Category, Category)]
-        public async Task Venv_With_SystemSitePackages_Ignores_Ambient_PYTHONNOUSERSITE()
+        public async Task Venv_With_SystemSitePackages_OwnSiteCustomize_Still_Runs()
         {
-            // Regression test: a customer's own leftover PYTHONNOUSERSITE=1 (e.g. a manual
-            // workaround predating this fix, exactly like the one mentioned in the original
-            // ticket) sitting in the *ambient* environment used to leak straight into the spawned
-            // host via ProcessStartInfo.EnvironmentVariables (which starts as a copy of this
-            // process's own environment), silently defeating a --system-site-packages venv's
-            // intent to leave user-site enabled — regardless of what our own code did or didn't
-            // set. Controller.ClearUserSiteEnvironmentOverride exists specifically to guarantee
-            // this venv flag decides the outcome, not whatever's ambient on the machine.
+            // Regression test: for a --system-site-packages venv, SetNoSiteFlag isn't set, so
+            // site.main() already ran during PythonEngine.Initialize() — before this fix, that
+            // meant it could already have imported and cached "sitecustomize" from the user site
+            // (enabled here via PYTHONUSERBASE, same as the ambient-PYTHONNOUSERSITE test below)
+            // before the venv's own site-packages, added afterwards in
+            // PostInitializationVenvSetup, ever got a chance to be searched. Without popping that
+            // cached module first, the venv's own sitecustomize.py would never run at all — only
+            // the user site's copy would. This asserts the venv's own copy does run.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            var venvSitePackages = Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"),
+                $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
+
+            var markerPath = _markerFile.Replace("\\", "/");
+            File.WriteAllText(Path.Combine(venvSitePackages, "sitecustomize.py"),
+                $"import codecs{Environment.NewLine}codecs.open('{markerPath}', 'a', encoding='utf-8').write('venv-sitecustomize\\n'){Environment.NewLine}");
+
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, EmbeddedPythonRuntimeBootstrap.UserSiteVersionFolder, "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(userSitePackages, "sitecustomize.py"),
+                $"import codecs{Environment.NewLine}codecs.open('{markerPath}', 'a', encoding='utf-8').write('usersite-sitecustomize\\n'){Environment.NewLine}");
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
+            Assert.Contains("venv-sitecustomize", markerLines);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_With_SystemSitePackages_Honors_Ambient_PYTHONNOUSERSITE()
+        {
+            // A --system-site-packages venv's own flag only says "also add the base install's
+            // site-packages" — it says nothing about user-site, which native CPython computes
+            // independently from PYTHONNOUSERSITE regardless of --system-site-packages. So an
+            // ambient PYTHONNOUSERSITE=1 (e.g. an administrator's own workaround, exactly like the
+            // one mentioned in the original ticket) must still suppress user-site here, matching
+            // what a normally-activated --system-site-packages venv would do. ProcessStartInfo
+            // already starts as a copy of this process's own environment, so this just needs
+            // nothing in our own code to actively defeat it.
             Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
 
             Environment.SetEnvironmentVariable("PYTHONNOUSERSITE", "1");
@@ -264,7 +302,7 @@ namespace UiPath.Python.Tests
                 $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
             WriteMarkerPth(venvSitePackages, "venv");
 
-            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, UserSiteVersionFolder, "site-packages")).FullName;
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, EmbeddedPythonRuntimeBootstrap.UserSiteVersionFolder, "site-packages")).FullName;
             WriteMarkerPth(userSitePackages, "usersite");
             Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
 
@@ -281,7 +319,7 @@ namespace UiPath.Python.Tests
             var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
 
             Assert.Contains("venv", markerLines);
-            Assert.Contains("usersite", markerLines);
+            Assert.DoesNotContain("usersite", markerLines);
         }
 
         private async Task<string[]> GetSysPath()
--- a/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
+++ b/Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
@@ -1,0 +1,313 @@
+using System;
+using System.IO;
+using System.Threading;
+using System.Threading.Tasks;
+using UiPath.TestUtils;
+using Xunit;
+using Assert = Xunit.Assert;
+
+namespace UiPath.Python.Tests
+{
+    // Regression test for STUD-81085: pointing Python Scope's Path at a venv must stop the
+    // interpreter's PEP-370 user-site directory (%APPDATA%\Roaming\Python\PythonXY\site-packages)
+    // from being processed at all. pywin32 registers its native DLL search directory
+    // (os.add_dll_directory) from a .pth file in whichever site directory it's installed into;
+    // if the user-site copy still gets processed alongside the venv's own copy, a differently
+    // built pywin32 there collides with the venv's copy and surfaces as
+    // "DLL load failed while importing win32api: The specified procedure could not be found."
+    //
+    // This test doesn't need real pywin32 binaries: the bug is that the user-site .pth runs at
+    // all when a venv is configured, so a .pth marker is a faithful, hermetic reproduction of the
+    // exact mechanism pywin32 relies on. Runs out-of-process (the real production default) since
+    // the fix depends on an environment variable set in the parent process before the host spawns.
+    public class VenvUserSiteIsolationTests : IDisposable
+    {
+        private const string Category = "Python";
+
+        private static readonly string EmbeddedRuntimePath = EmbeddedPythonRuntimeBootstrap.EnsureRuntimePath();
+        private static readonly string EmbeddedLibraryPath = EmbeddedPythonRuntimeBootstrap.GetPythonLibraryPath(EmbeddedRuntimePath);
+
+        // Must match the running embeddable interpreter's sys.version_info (3.14.5) — Windows
+        // user-site resolves to <PYTHONUSERBASE>\Python<major><minor>\site-packages.
+        private const string UserSiteVersionFolder = "Python314";
+
+        private readonly string _rootDir;
+        private readonly string _venvDir;
+        private readonly string _userBaseDir;
+        private readonly string _markerFile;
+        private readonly string _previousUserBase;
+        private readonly string _previousNoUserSite;
+
+        public VenvUserSiteIsolationTests()
+        {
+            _rootDir = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), "venv-usersite-tests", Guid.NewGuid().ToString("N"))).FullName;
+            _venvDir = Path.Combine(_rootDir, ".venv");
+            _userBaseDir = Path.Combine(_rootDir, "userbase");
+            _markerFile = Path.Combine(_rootDir, "marker.txt");
+
+            _previousUserBase = Environment.GetEnvironmentVariable("PYTHONUSERBASE");
+            _previousNoUserSite = Environment.GetEnvironmentVariable("PYTHONNOUSERSITE");
+        }
+
+        public void Dispose()
+        {
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", _previousUserBase);
+            Environment.SetEnvironmentVariable("PYTHONNOUSERSITE", _previousNoUserSite);
+            try { Directory.Delete(_rootDir, true); } catch { /* best effort cleanup */ }
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_Does_Not_Process_UserSite_PthFiles()
+        {
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            var venvSitePackages = Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {EmbeddedRuntimePath}{Environment.NewLine}");
+            WriteMarkerPth(venvSitePackages, "venv");
+
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, UserSiteVersionFolder, "site-packages")).FullName;
+            WriteMarkerPth(userSitePackages, "usersite");
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile)
+                ? File.ReadAllLines(_markerFile)
+                : Array.Empty<string>();
+
+            // The venv's own .pth must still run — this isn't about disabling site processing,
+            // only about excluding the unrelated per-user directory.
+            Assert.Contains("venv", markerLines);
+            Assert.DoesNotContain("usersite", markerLines);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_With_SystemSitePackages_Still_Processes_UserSite_PthFiles()
+        {
+            // Mirrors CPython's own site.py venv() function: it only forces
+            // ENABLE_USER_SITE = False when the venv was created *without*
+            // --system-site-packages. A real, natively-activated --system-site-packages venv
+            // leaves user-site enabled (the normal, non-venv computation applies instead) — so
+            // this fix must not suppress it there either, or it would diverge from native parity
+            // for a case that already accepts the same DLL-collision exposure as any other
+            // non-venv interpreter.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            var venvSitePackages = Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"),
+                $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
+            WriteMarkerPth(venvSitePackages, "venv");
+
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, UserSiteVersionFolder, "site-packages")).FullName;
+            WriteMarkerPth(userSitePackages, "usersite");
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile)
+                ? File.ReadAllLines(_markerFile)
+                : Array.Empty<string>();
+
+            Assert.Contains("venv", markerLines);
+            Assert.Contains("usersite", markerLines);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_PythonHome_Resolves_To_Declared_Base_Install()
+        {
+            // Regression test for a separate bug found while exercising this fix: pointing
+            // PythonHome directly at the venv root (rather than the base install its own
+            // pyvenv.cfg declares) makes native init fail outright once a real installer-based
+            // Python is used, since a venv has no standard library of its own. The embeddable test
+            // runtime's own ._pth-based bootstrap resolves its stdlib independently of PythonHome,
+            // so it can't catch that failure directly — but PythonHome still governs what the
+            // interpreter reports as sys.base_prefix regardless, which is what this asserts.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages"));
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {EmbeddedRuntimePath}{Environment.NewLine}");
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            string basePrefix;
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+
+                var script = await engine.LoadScript(
+                    "import sys\ndef check():\n    return sys.base_prefix\n",
+                    CancellationToken.None);
+                var result = await engine.InvokeMethod(script, "check", null, CancellationToken.None);
+                basePrefix = (string)engine.Convert(result, typeof(string));
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            Assert.Equal(
+                Path.GetFullPath(EmbeddedRuntimePath).TrimEnd(Path.DirectorySeparatorChar),
+                Path.GetFullPath(basePrefix).TrimEnd(Path.DirectorySeparatorChar),
+                ignoreCase: true);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_Does_Not_Leak_BaseInstall_SitePackages()
+        {
+            // Regression test for a second, distinct leak path found by inspecting a real venv's
+            // actual sys.path: CPython's site.main() unconditionally adds the base install's own
+            // site-packages (and the bare base install prefix itself) unless something narrows
+            // PREFIXES first — which never happened for the embedded interpreter, since site.py's
+            // own venv() detection can't trigger for it. PYTHONNOUSERSITE never covered this; only
+            // SetNoSiteFlag (skipping site.main() entirely for this case) does.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages"));
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {EmbeddedRuntimePath}{Environment.NewLine}");
+
+            var sysPath = await GetSysPath();
+            var normalizedBase = Path.GetFullPath(EmbeddedRuntimePath).TrimEnd(Path.DirectorySeparatorChar);
+
+            Assert.DoesNotContain(sysPath, p => Path.GetFullPath(p).TrimEnd(Path.DirectorySeparatorChar)
+                .Equals(normalizedBase, StringComparison.OrdinalIgnoreCase));
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_With_SystemSitePackages_Still_Includes_BaseInstall_SitePackages()
+        {
+            // The --system-site-packages case must keep behaving exactly as it did before this
+            // change: SetNoSiteFlag is only set for the default (ShouldDisableUserSite) case, so
+            // site.main() still runs normally here and still adds the base install unconditionally
+            // — which happens to already be correct for this specific flag.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages"));
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"),
+                $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
+
+            var sysPath = await GetSysPath();
+            var normalizedBase = Path.GetFullPath(EmbeddedRuntimePath).TrimEnd(Path.DirectorySeparatorChar);
+
+            Assert.Contains(sysPath, p => Path.GetFullPath(p).TrimEnd(Path.DirectorySeparatorChar)
+                .Equals(normalizedBase, StringComparison.OrdinalIgnoreCase));
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_SiteCustomize_Still_Runs()
+        {
+            // SetNoSiteFlag skips site.main() entirely for the default venv case, which would also
+            // silently skip sitecustomize.py auto-import (some environments rely on it for
+            // corporate setup) unless something restores it — Engine.PostInitializationVenvSetup
+            // explicitly calls site.execsitecustomize() for exactly this reason.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            var venvSitePackages = Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"), $"home = {EmbeddedRuntimePath}{Environment.NewLine}");
+
+            var markerPath = _markerFile.Replace("\\", "/");
+            File.WriteAllText(Path.Combine(venvSitePackages, "sitecustomize.py"),
+                $"import codecs{Environment.NewLine}codecs.open('{markerPath}', 'a', encoding='utf-8').write('sitecustomize\\n'){Environment.NewLine}");
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
+            Assert.Contains("sitecustomize", markerLines);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_With_SystemSitePackages_Ignores_Ambient_PYTHONNOUSERSITE()
+        {
+            // Regression test: a customer's own leftover PYTHONNOUSERSITE=1 (e.g. a manual
+            // workaround predating this fix, exactly like the one mentioned in the original
+            // ticket) sitting in the *ambient* environment used to leak straight into the spawned
+            // host via ProcessStartInfo.EnvironmentVariables (which starts as a copy of this
+            // process's own environment), silently defeating a --system-site-packages venv's
+            // intent to leave user-site enabled — regardless of what our own code did or didn't
+            // set. Controller.ClearUserSiteEnvironmentOverride exists specifically to guarantee
+            // this venv flag decides the outcome, not whatever's ambient on the machine.
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            Environment.SetEnvironmentVariable("PYTHONNOUSERSITE", "1");
+
+            var venvSitePackages = Directory.CreateDirectory(Path.Combine(_venvDir, "Lib", "site-packages")).FullName;
+            File.WriteAllText(Path.Combine(_venvDir, "pyvenv.cfg"),
+                $"home = {EmbeddedRuntimePath}{Environment.NewLine}include-system-site-packages = true{Environment.NewLine}");
+            WriteMarkerPth(venvSitePackages, "venv");
+
+            var userSitePackages = Directory.CreateDirectory(Path.Combine(_userBaseDir, UserSiteVersionFolder, "site-packages")).FullName;
+            WriteMarkerPth(userSitePackages, "usersite");
+            Environment.SetEnvironmentVariable("PYTHONUSERBASE", _userBaseDir);
+
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+            }
+            finally
+            {
+                await engine.Release();
+            }
+
+            var markerLines = File.Exists(_markerFile) ? File.ReadAllLines(_markerFile) : Array.Empty<string>();
+
+            Assert.Contains("venv", markerLines);
+            Assert.Contains("usersite", markerLines);
+        }
+
+        private async Task<string[]> GetSysPath()
+        {
+            var engine = EngineProvider.Get(Version.Python_310, _venvDir, EmbeddedLibraryPath, inProcess: false);
+            try
+            {
+                await engine.Initialize(null, CancellationToken.None, 60);
+
+                var script = await engine.LoadScript(
+                    "import sys\ndef check():\n    return list(sys.path)\n",
+                    CancellationToken.None);
+                var result = await engine.InvokeMethod(script, "check", null, CancellationToken.None);
+                return (string[])engine.Convert(result, typeof(string[]));
+            }
+            finally
+            {
+                await engine.Release();
+            }
+        }
+
+        private void WriteMarkerPth(string siteDir, string tag)
+        {
+            var markerPath = _markerFile.Replace("\\", "/");
+            var pthLine = $"import codecs; codecs.open('{markerPath}', 'a', encoding='utf-8').write('{tag}\\n')";
+            File.WriteAllText(Path.Combine(siteDir, $"zzz_{tag}_marker.pth"), pthLine + Environment.NewLine);
+        }
+    }
+}
--- a/Activities/Python/UiPath.Python.Tests/VenvVersionValidationTests.cs
+++ b/Activities/Python/UiPath.Python.Tests/VenvVersionValidationTests.cs
@@ -1,0 +1,79 @@
+using System;
+using System.IO;
+using UiPath.TestUtils;
+using Xunit;
+using Assert = Xunit.Assert;
+
+namespace UiPath.Python.Tests
+{
+    // Regression tests for EngineProvider.ValidateVenvVersion (called from the public
+    // ValidateInstallation): a venv's declared Python version (its own pyvenv.cfg) must match the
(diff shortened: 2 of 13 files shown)


alexandru-petre (commented):
## Code review — findings

**Verdict: NEEDS CHANGES** (1 HIGH, 8 MEDIUM, 2 LOW). Reviewed all 11 changed files against STUD=81085 (description + all 7 comments) and the PR description. 12 inline findings below.

### Highlights

- **HIGH** — the `PythonHome` fix does not cover Microsoft Store Python venvs. `pyvenv.cfg`'s `home` there points at the app-execution-alias folder, which contains only reparse-point `.exe` stubs — no `Lib`, no `Lib\encodings`. Verified locally against a stdlib-`venv` config with `home = ...\WindowsApps\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0`. Not a regression (the old `PythonHome = _path` was equally broken), but the headline fix does not close a mainstream Windows install shape.
- The new venv version cross-check silently no-ops for `virtualenv`/`uv` venvs, which write `version_info` rather than `version`.
- `ClearUserSiteEnvironmentOverride` contradicts its own native-parity rationale — a natively-launched interpreter *does* honour `PYTHONNOUSERSITE=1`.

### On the failed quality gate

Pulled the 34 issues from the SonarCloud API for this PR. **D Security and D Reliability are driven entirely by pre-existing code in the touched files**, not by this diff: `S4790` (MD5 in `Engine.Hash`) and `S6444` (regex without timeout in `EngineProvider`) for Security; `S2551` (`lock(this)` x3) and `S2445` for Reliability. The only genuinely PR-caused Sonar issue is `S3776` on `Engine.Initialize` (complexity 29 vs. limit 15), commented inline. Suggest a separate PR or a gate exception rather than scope creep here.

### Evidence gap

The ticket's reproduction package (`Test-Pywin32VenvRepro (2).zip`, attached 2026-08-20) postdates this branch's only commit (`3b2a9d61`, 2026-08-19). Nothing in the PR records a run against it — see the inline comment on `VenvUserSiteIsolationTests.cs`.

### Done well

- The `SetNoSiteFlag` block documents *why* the env-var approach was abandoned and why call ordering matters — exactly the reasoning that otherwise gets re-litigated six months later.
- Replacing the unbounded `GetVenvPath(maxLevels: 3)` ancestor walk with root-or-`Scripts`/`bin` detection is the right narrowing, and both false-positive shapes are pinned by tests.
- Both sides of the `--system-site-packages` fork assert opposite outcomes from the same code path — the pairing that actually proves the flag is load-bearing.
- Catching the second-order `sitecustomize.py` loss and pinning it with `Venv_SiteCustomize_Still_Runs` was sharp.
- Clean migration hygiene: zero surviving references to the removed `IsVenv`/`GetVenvPath`. Localization convention followed — only the canonical English resx was touched.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -55,20 +55,71 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
-                            if (!_path.IsNullOrEmpty())
-                                PythonEngine.PythonHome = _path;
+                            // A venv's own folder is not a valid PythonHome: it only has
+                            // Lib\site-packages, not the standard library (Lib\encodings etc.), so
+                            // pointing the native interpreter at it fails at the very first import
+                            // with "Fatal Python error: Failed to import encodings module". Use the
+                            // base install recorded in the venv's own pyvenv.cfg instead — exactly
+                            // what a normally-activated venv resolves to on its own. EngineProvider
+                            // already validated that base install's version matches _libraryPath's.
+                            var pythonHome = venv != null ? ResolvePythonHome(venv.Home) : _path;
```
alexandru-petre: **[Correctness/HIGH] `PythonHome` fix does not cover Microsoft Store Python venvs**

`ResolvePythonHome(venv.Home)` trusts `pyvenv.cfg`'s `home` to be a real install prefix. For a venv created from a Microsoft-Store-installed Python, `home` points at the app-execution-alias folder, which contains only reparse-point `.exe` stubs — no `Lib`, no `Lib\encodings`, no `pythonXY.dll`.

Verified concretely against a stdlib-`venv` config:

```
home = C:\Users\<user>\AppData\Local\Microsoft\WindowsApps\PythonSoftwareFoundation.Python.3.13_qbz5n2kfra8p0
version = 3.13.14
```

That directory holds `python.exe`, `pip.exe`, `idle.exe` and nothing else. `Lib` does not exist.

**Why it matters:** `PythonHome` is then set to a directory with no standard library, producing exactly the `Fatal Python error: Failed to import encodings module` that the comment directly above says this change prevents. `SetNoSiteFlag()` also removes `site.py` as a fallback path source. This is *not* a regression — the old `PythonHome = _path` was equally broken — but it means the PR's headline fix does not close the case for a mainstream Windows install shape.

**Suggested fix:** after computing `pythonHome`, verify it actually carries a stdlib (Windows: `Lib\encodings` exists; POSIX: `lib/python*/encodings`). When it does not, fall back to `Path.GetDirectoryName(_libraryPath)` — `LibraryPath` is mandatory and validated in `PythonScope.cs:152-154`, so its directory is by construction the real install root. Worth a `VenvUserSiteIsolationTests` case for a `pyvenv.cfg` whose `home` has no `Lib\encodings`.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
```diff
@@ -0,0 +1,84 @@
...
+                return null;
+
+            var kv = new Dictionary<string, string>(StringComparer.OrdinalIgnoreCase);
+            foreach (var line in File.ReadLines(cfgFile))
+            {
+                var parts = line.Split('=', 2);
+                if (parts.Length == 2)
+                    kv[parts[0].Trim()] = parts[1].Trim();
+            }
+
```
alexandru-petre: **[Correctness/MEDIUM] Venv version cross-check silently no-ops for `virtualenv`/`uv` venvs**

Only the `version` key is read. CPython's stdlib `venv` writes `version`, but `virtualenv` and `uv` write **`version_info`** instead (e.g. `version_info = 3.10.4.final.0`).

For venvs created by either tool `venv.Version` is `null`, so `EngineProvider.ValidateVenvVersion` returns early and the new cross-check never fires — the mismatched-ABI case it exists to catch still reaches the engine and fails with the confusing deep error the check was added to replace. The skip is silent; nothing logs that validation did not run.

Detection itself is unaffected (those tools do write `home`), so this only degrades the new validation.

**Suggested fix:** fall back to `version_info` when `version` is absent —

```csharp
kv.TryGetValue("version", out var version);
if (version == null)
    kv.TryGetValue("version_info", out version);
```

`TryParseVenvVersion` already tolerates the extra `.final.0` segments since it takes only the first two parts. Worth a `VenvDetectionTests` case for a `version_info`-only config. Please confirm the key names against a locally created `virtualenv`/`uv` venv before applying.

Activities/Python/UiPath.Python/Impl/OutOfProcessEngine.cs
```diff
@@ -50,11 +50,29 @@ public Task Initialize(string workingFolder, CancellationToken ct, double timeou
...
+            // fix), it would otherwise leak into the host and silently force user-site off anyway,
+            // regardless of what SetNoSiteFlag does or doesn't do for the other case. Clearing it
+            // explicitly for the child guarantees the venv's own IncludeSystemSitePackages flag is
+            // what decides this, not whatever's ambient on the machine.
             _provider = new Controller<IPythonService>()
             {
                 PythonHostLibFile = ServiceDll_x64,
-                Visible = _visible
+                Visible = _visible,
+                ClearUserSiteEnvironmentOverride = venv != null && venv.IncludeSystemSitePackages
```
alexandru-petre: **[Correctness/MEDIUM] `ClearUserSiteEnvironmentOverride` contradicts its own native-parity rationale**

For a `--system-site-packages` venv the host's `PYTHONNOUSERSITE` is unconditionally removed. The rationale a few lines up says the intent is to "leave user-site exactly as a normal, non-embedded interpreter would" — but a normal, natively-launched interpreter **does** honour `PYTHONNOUSERSITE=1`. Stripping it is a *divergence* from native parity, not a restoration of it.

An operator who intentionally sets `PYTHONNOUSERSITE=1` machine-wide silently loses it after upgrading — including the exact workaround this ticket's customer was told to apply (Jira comment 4313615). `Venv_With_SystemSitePackages_Ignores_Ambient_PYTHONNOUSERSITE` locks the behaviour in as intended, so the code and its own justifying comment disagree about what the goal is.

**Suggested fix:** pick one and make code, comment and test agree.
- If native parity is the goal → drop `ClearUserSiteEnvironmentOverride` and let the ambient variable through.
- If overriding is intended → rewrite the comment to say so explicitly (*the venv flag deliberately wins over ambient configuration*) and rename the test to match.

Activities/Python/UiPath.Python.Tests/VenvVersionValidationTests.cs
```diff
@@ -0,0 +1,79 @@
...
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public void MatchingVenvVersion_DoesNotThrow()
+        {
+            Skip.IfNot(Directory.Exists(EmbeddedRuntimePath));
+
+            WriteVenvCfg("3.14.5");
+
+            // Must not throw.
+            EngineProvider.ValidateInstallation(_venvDir, EmbeddedLibraryPath);
```
alexandru-petre: **[Correctness/MEDIUM] Test contains no assertion**

`MatchingVenvVersion_DoesNotThrow` (here) and `NonVenvPath_DoesNotThrow` (line 76) both call `EngineProvider.ValidateInstallation(...)` and end. Neither body contains an `Assert` call, so the intent is carried only by a comment and the method name — a future reader cannot distinguish a deliberate no-throw assertion from an unfinished test.

**Suggested fix**, matching the shape already used in `ValidatePythonLibraryTests.cs:101-103`:

```csharp
Assert.Null(Record.Exception(() => EngineProvider.ValidateInstallation(_venvDir, EmbeddedLibraryPath)));
```

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
```diff
@@ -0,0 +1,313 @@
...
+                ? File.ReadAllLines(_markerFile)
+                : Array.Empty<string>();
+
+            Assert.Contains("venv", markerLines);
+            Assert.Contains("usersite", markerLines);
+        }
+
+        [Fact]
+        [Trait(TestCategories.Category, Category)]
+        public async Task Venv_PythonHome_Resolves_To_Declared_Base_Install()
```
alexandru-petre, line 141: **[Correctness/MEDIUM] No regression coverage against an installer-based Python**

All seven new engine-level tests run against the embeddable `python-3.14.5-embed-amd64` distribution. This test's own comment (lines 138-143) concedes that runtime "resolves its stdlib independently of PythonHome, so it can't catch that failure directly", and substitutes a weaker `sys.base_prefix` assertion.

So the PR's headline `PythonHome` change — the one thing that hard-fails a real deployment if wrong — has no test that can actually fail when it regresses. The reported bug is Python 3.10 + installer-based Python + real pywin32; the suite exercises 3.14 + embeddable + `.pth` markers. The HIGH finding on `Engine.cs:84` is a direct consequence of this gap.

Also worth noting: the ticket's repro package (`Test-Pywin32VenvRepro (2).zip`, attached 2026-08-20) postdates this branch's only commit (2026-08-19).

**Suggested fix:** record in the PR body a manual run against that repro on a real installer-based Python 3.10 venv, with pywin32 installed in *both* the venv and `%APPDATA%\Roaming\Python\Python310\site-packages`. If a CI-runnable form is wanted, add a `Skip.IfNot`-gated test keyed on an installer-based Python being present on the agent.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
```diff
@@ -0,0 +1,84 @@
...
+        /// <summary>
+        /// Parses pyvenv.cfg at <paramref name="path"/>, if present. Requiring at least one of the
+        /// keys a real venv config always has (home/version) avoids treating an unrelated file that
+        /// merely happens to be named pyvenv.cfg as a venv.
+        /// </summary>
+        private static VenvInfo TryReadVenvConfig(string path)
+        {
+            var cfgFile = Path.Combine(path, "pyvenv.cfg");
+            if (!File.Exists(cfgFile))
+                return null;
```
alexandru-petre, line 41: **[Bugs/MEDIUM] An unreadable `pyvenv.cfg` now hard-fails the scope**

`File.ReadLines(cfgFile)` is unguarded. The removed `Engine.IsVenv` only called `File.Exists`, which never throws. `GetVenvInfo` is now reached from three points on the execution path — `EngineProvider.ValidateInstallation` (called unguarded from `PythonScope.cs:157`), `OutOfProcessEngine.Initialize:53`, and `Engine.Initialize:69`.

A `pyvenv.cfg` that exists but cannot be read — ACLs on a locked-down robot account, an AV sharing violation, a concurrent `pip` rewriting it — now throws a raw `IOException`/`UnauthorizedAccessException` out of `ValidateInstallation` and fails the whole activity.

This breaks the convention every neighbouring validator follows: `ValidatePythonLibrary` (`EngineProvider.cs:303-306`) and `ValidatePythonExecutable` (`EngineProvider.cs:265-269`) both deliberately swallow and let downstream produce a precise error.

**Suggested fix:** wrap the read in `try { ... } catch { return null; }` inside `TryReadVenvConfig`, matching the "unable to inspect — let engine initialization produce its own error" style already used in this call chain.

Activities/Python/UiPath.Python/EngineProvider.cs
```diff
@@ -142,13 +142,51 @@ private static bool Autodetect(string pythonFullPath, out Version version, out E
...
+        {
+            var venv = VenvDetection.GetVenvInfo(path);
+            if (venv?.Version == null || !TryParseVenvVersion(venv.Version, out int venvMajor, out int venvMinor))
+                return;
+
+            if (!TryGetLibraryVersion(libraryPath, out int libMajor, out int libMinor))
+                return;
+
+            if (venvMajor != libMajor || venvMinor != libMinor)
+                throw new NotSupportedException(
```
alexandru-petre, line 176: **[BackCompat/MEDIUM] New hard failure on a public API, not called out in the PR description**

`ValidateInstallation` is `public` and now throws `NotSupportedException` whenever a venv's declared version differs from `LibraryPath`'s. It is called from `PythonScope.ExecuteAsync` (`PythonScope.cs:157`) and from `PythonService` (`PythonService.cs:64`).

A workflow whose venv contains only pure-Python packages, and whose declared version merely differs from `LibraryPath`, runs today and hard-fails after the upgrade. That is defensible — the pairing is genuinely unsound — but it is a silent behaviour change on a public API shipping in a **patch** release.

**Suggested fix:** no code change needed if the stricter behaviour is intended; add an explicit "Behaviour changes" line to the PR description naming the new `NotSupportedException`, so it carries into the patch release notes.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -55,20 +55,71 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
                             Trace.TraceInformation($"Initializing Python runtime using version {_version} and path {_path}");
                             Stopwatch sw = Stopwatch.StartNew();
 
+                            // Detected before Initialize(): a real venv (activated normally, e.g.
```
alexandru-petre: **[Maintainability/MEDIUM] `Engine.Initialize` cognitive complexity pushed to 29**

The method grew from ~30 to ~105 lines, of which roughly 60 are comment prose (lines 58-68 and 88-114). Sonar raises `csharpsquid:S3776` at CRITICAL — cognitive complexity 29 against a limit of 15. This is the **only** genuinely PR-caused Sonar issue on the diff.

Venv detection, `PythonHome` resolution and the no-site decision are three separable concerns wedged inside a double-checked lock inside a `try`. The rationale itself is valuable — at this density it just buries the eight lines of actual logic.

**Suggested fix:** extract a private `ConfigureRuntime(VenvDetection.VenvInfo venv)` covering `SetDllDirectory` / `Runtime.PythonDLL` / `PythonHome` / `SetNoSiteFlag`, and move the two long explanatory blocks onto that method's XML doc. `Initialize` then reads as detect → configure → initialize → post-setup.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -55,20 +55,71 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
                                 Runtime.PythonDLL = _libraryPath;
 
-                            if (!_path.IsNullOrEmpty())
-                                PythonEngine.PythonHome = _path;
+                            // A venv's own folder is not a valid PythonHome: it only has
+                            // Lib\site-packages, not the standard library (Lib\encodings etc.), so
+                            // pointing the native interpreter at it fails at the very first import
+                            // with "Fatal Python error: Failed to import encodings module". Use the
+                            // base install recorded in the venv's own pyvenv.cfg instead — exactly
+                            // what a normally-activated venv resolves to on its own. EngineProvider
```
alexandru-petre: **[Maintainability/MEDIUM] This comment claims a validation that does not exist**

> EngineProvider already validated that base install's version matches `_libraryPath`'s.

`EngineProvider.ValidateVenvVersion` compares `pyvenv.cfg`'s `version` **key** against `LibraryPath`. It never inspects the install at `home`.

The claim is load-bearing — it is the stated reason `ResolvePythonHome(venv.Home)` needs no validation of its own. A reader tracing the `Engine.cs:84` HIGH finding will trust a guarantee that is not there. And per the `VenvDetection.cs:47` comment, the `version` key is frequently absent (`virtualenv`/`uv`), so even the weaker check often does not run at all.

**Suggested fix:** either restate accurately — *"EngineProvider cross-checked the version declared in pyvenv.cfg against `_libraryPath`; the install at `home` is not itself validated"* — or add the missing validation per the `Engine.cs:84` fix and update the comment to match.

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
```diff
@@ -0,0 +1,313 @@
...
+    public class VenvUserSiteIsolationTests : IDisposable
+    {
+        private const string Category = "Python";
+
+        private static readonly string EmbeddedRuntimePath = EmbeddedPythonRuntimeBootstrap.EnsureRuntimePath();
+        private static readonly string EmbeddedLibraryPath = EmbeddedPythonRuntimeBootstrap.GetPythonLibraryPath(EmbeddedRuntimePath);
+
+        // Must match the running embeddable interpreter's sys.version_info (3.14.5) — Windows
+        // user-site resolves to <PYTHONUSERBASE>\Python<major><minor>\site-packages.
+        private const string UserSiteVersionFolder = "Python314";
```
alexandru-petre: **[Maintainability/MEDIUM] Embedded-runtime version hard-coded and duplicated across test files**

`UserSiteVersionFolder = "Python314"` here, and `WriteVenvCfg("3.14.5")` in `VenvVersionValidationTests.cs:60`, each restate `EmbeddedPythonRuntimeBootstrap.PythonVersion` — which is a `private const` (`EmbeddedPythonRuntimeBootstrap.cs:13`) and so cannot be referenced.

When the embedded runtime is bumped, `Venv_Does_Not_Process_UserSite_PthFiles` keeps **passing but becomes vacuous**: its `Assert.DoesNotContain("usersite", ...)` succeeds because the user-site folder name no longer resolves, not because isolation works. The drift is caught only indirectly, by sibling tests failing.

**Suggested fix:** make `EmbeddedPythonRuntimeBootstrap.PythonVersion` `public` (or expose `public static string UserSiteVersionFolder => $"Python{major}{minor}"`) and derive both constants from it.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -55,20 +55,71 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
+                            // already validated that base install's version matches _libraryPath's.
+                            var pythonHome = venv != null ? ResolvePythonHome(venv.Home) : _path;
+                            if (!pythonHome.IsNullOrEmpty())
```
alexandru-petre: **[Correctness/LOW] `PythonHome` silently left unset when `pyvenv.cfg` has `version` but no `home`**

`TryReadVenvConfig` accepts a config carrying either `home` **or** `version` (`VenvDetection.cs:44`). With `version` only, `venv.Home` is `null`, `ResolvePythonHome` returns `null`, and this `IsNullOrEmpty` guard skips setting `PythonHome` altogether.

Before this change `PythonHome` was always assigned (`_path`). The new path leaves it at pythonnet's default, so the interpreter falls back to registry/`._pth` discovery — a third behaviour that neither the code nor the comments describe.

**Suggested fix:** either require `home` for a config to count as a venv, or fall back to `_path` when `ResolvePythonHome` yields `null`, so the assignment is never skipped silently.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -55,20 +55,71 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
+                            // site.addsitedir() call in PostInitializationVenvSetup for the venv's own
+                            // site-packages, and the site.execsitecustomize() call there preserving
+                            // sitecustomize.py support, both keep working. Must come after
+                            // Runtime.PythonDLL/PythonHome are set, not before — calling it earlier
+                            // left Runtime.PythonDLL null by the time the host tried to use it (and,
+                            // per a known pythonnet issue, SetNoSiteFlag itself can be silently
+                            // ignored on Windows unless another PythonEngine call already preceded
+                            // it — PythonHome, set just above, already satisfies that).
+                            if (venv != null && venv.ShouldDisableUserSite)
+                                PythonEngine.SetNoSiteFlag();
```
alexandru-petre: **[Maintainability/LOW] `Py_NoSiteFlag` is deprecated since CPython 3.12, removal planned for 3.15**

`PythonEngine.SetNoSiteFlag()` writes CPython's legacy global `Py_NoSiteFlag`, deprecated in 3.12 with removal announced for 3.15. `VersionExtensions._supportedRuntimeVersions` (`Version.cs:167`) already lists up to 3.14, so this works for every currently supported version — no action needed today.

The risk is the failure mode when 3.15 is added: because the fix is "prevent, don't clean up after", the isolation would silently stop taking effect and the original pywin32 DLL mismatch would return, rather than surfacing an error.

**Suggested fix:** note the 3.15 removal at this call site, and add "confirm `Py_NoSiteFlag` still applies" to the checklist for extending `_supportedRuntimeVersions`. The forward-looking replacement is `PyConfig.site_import` via `Py_InitializeFromConfig`.

Activities/Python/UiPath.Python/Impl/OutOfProcessEngine.cs
alexandru-petre: **Consolidating** — Copilot independently flagged this same line, and its reasoning settles the question I left open above. The direction is now clear: **preserve the inherited variable, don't strip it.**

Traced through CPython's `site.py` to confirm the claim rather than take it at face value:

- `venv()` sets `ENABLE_USER_SITE = False` **only** when `include-system-site-packages` is false. When it is true, `ENABLE_USER_SITE` is left at `None`.
- `main()` then hits `if ENABLE_USER_SITE is None: ENABLE_USER_SITE = check_enableusersite()`.
- `check_enableusersite()` returns `False` when `sys.flags.no_user_site` is set — which is exactly what `PYTHONNOUSERSITE` sets.

So a natively-launched `--system-site-packages` venv **does** honour `PYTHONNOUSERSITE=1`. The flag permits system packages; it does not override the independent user-site setting. Stripping the variable is therefore a divergence from native parity rather than a restoration of it — which makes the comment at lines 62-70 the accurate statement of intent, and the code the thing that is wrong.

**Concrete fix:**

1. Drop `ClearUserSiteEnvironmentOverride` from this initializer, and remove the property plus the `psi.EnvironmentVariables.Remove("PYTHONNOUSERSITE")` call from `Controller.cs`.
2. Trim the comment block at lines 62-70 accordingly — the parity rationale survives, the plumbing it justified does not.
3. Revise `Venv_With_SystemSitePackages_Ignores_Ambient_PYTHONNOUSERSITE`. It currently pins the wrong behaviour. Inverted it becomes genuinely useful: ambient `PYTHONNOUSERSITE=1` + a `--system-site-packages` venv should yield *no* `usersite` marker.

`SetNoSiteFlag()` still covers the default-venv case unchanged, so the STUD=81085 fix itself is unaffected by this.

Copilot's thread on this same line has been resolved into this one.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
```diff
@@ -0,0 +1,84 @@
...
+                var parts = line.Split('=', 2);
+                if (parts.Length == 2)
+                    kv[parts[0].Trim()] = parts[1].Trim();
+            }
+
+            if (!kv.ContainsKey("home") && !kv.ContainsKey("version"))
+                return null;
+
+            kv.TryGetValue("home", out var home);
+            kv.TryGetValue("version", out var version);
```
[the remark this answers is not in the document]
alexandru-petre: Duplicate — same finding and same fix as https://github.com/UiPath/Community.Activities/pull/596#discussion_r3852970944, posted a minute earlier. Reached independently, which is reasonable corroboration for the `version_info` claim itself; tracking on that thread.

One correction so this doesn't send anyone down a blind alley: *"This issue also appears on line 81 of the same file"* is not right. Line 81 is `return TryReadVenvConfig(Path.GetDirectoryName(path));` — nothing to do with version parsing. That line does have a separate, real defect (trailing-separator handling), which this same review raised as its suppressed comment.

Resolving as a duplicate.

Activities/Python/UiPath.Python/Impl/OutOfProcessEngine.cs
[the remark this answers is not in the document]
alexandru-petre: Agreed — and verified against CPython's `site.py` rather than taken at face value: `venv()` only forces `ENABLE_USER_SITE = False` when `include-system-site-packages` is false, so in this mode `main()` falls through to `check_enableusersite()`, which honours `sys.flags.no_user_site`. Your reading is correct, and it settles an open question I had left on this same line.

Consolidated into https://github.com/UiPath/Community.Activities/pull/596#discussion_r3852970950, which carries the full trace and the concrete fix (drop the property and the `Controller.cs` plumbing, invert the regression test).

Resolving here so this stays on one thread.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
```diff
@@ -0,0 +1,84 @@
...
+
+            var direct = TryReadVenvConfig(path);
+            if (direct != null)
+                return direct;
+
+            var folderName = Path.GetFileName(path.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar));
+            if (!Array.Exists(VenvBinFolderNames, name => string.Equals(name, folderName, StringComparison.OrdinalIgnoreCase)))
+                return null;
+
+            return TryReadVenvConfig(Path.GetDirectoryName(path));
```
alexandru-petre: **[Bugs/MEDIUM] A trailing separator on `Path` silently disables launcher-folder detection**

Surfacing this one deliberately: Copilot raised it in the **suppressed** section of its review, so it is collapsed and easy to miss. I verified it empirically and it is real — above the bar for suppression, in my view.

`folderName` is derived from the *trimmed* path, but `Path.GetDirectoryName` on line 81 is handed the *raw* one. Measured on .NET:

| `path` | `GetDirectoryName(path)` | `GetFileName(trimmed)` |
|---|---|---|
| `C:\venv\Scripts\` | `C:\venv\Scripts` | `Scripts` |
| `C:\venv\Scripts` | `C:\venv` | `Scripts` |
| `C:\venv\Scripts/` | `C:\venv\Scripts` | `Scripts` |

So for `<venv>\Scripts\` the guard on line 78 **passes** (`folderName` is `Scripts`), execution reaches line 81, and `TryReadVenvConfig` then looks for `pyvenv.cfg` *inside the `Scripts` folder*. It is not there, so `GetVenvInfo` returns `null`.

**Why it matters:** the failure is silent and total. A `null` here means `venv` is `null` in both `Engine.Initialize` and `OutOfProcessEngine.Initialize`, so `SetNoSiteFlag()` never fires, `PythonHome` falls back to `_path`, and `PostInitializationVenvSetup` no-ops — the entire STUD=81085 fix is skipped, and the user gets the original pywin32 DLL mismatch back with no diagnostic. `Path` is a free-text property on the activity, so a trailing separator is an ordinary thing for a user to type.

Scope is narrow — venv-*root* paths are unaffected, since `Path.Combine` handles a trailing separator fine, and this only bites the `Scripts`/`bin` form. But the fix is one line: hoist the trimmed value and use it for both.

```csharp
var trimmed = path.TrimEnd(Path.DirectorySeparatorChar, Path.AltDirectorySeparatorChar);
var folderName = Path.GetFileName(trimmed);
if (!Array.Exists(VenvBinFolderNames, name => string.Equals(name, folderName, StringComparison.OrdinalIgnoreCase)))
    return null;

return TryReadVenvConfig(Path.GetDirectoryName(trimmed));
```

Worth a `VenvDetectionTests` case too — `LauncherSubfolder_Is_Detected` currently only feeds paths built by `Directory.CreateDirectory(...).FullName`, which never carry a trailing separator, so nothing in the suite exercises this shape. An `[InlineData]` row with a trailing `\` would have caught it.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -296,29 +354,35 @@ private static string GetEnvSitePackagesPath(string venvPath)
...
+
+                    // SetNoSiteFlag (see Initialize()) skips site.main() entirely for a default
+                    // venv, which also skips its sitecustomize.py auto-import — some environments
+                    // rely on that for corporate setup (proxies, logging, etc.), and it did run
+                    // today before this change, so preserve it explicitly. Safe to call even when
+                    // SetNoSiteFlag wasn't set (--system-site-packages venvs): site.main() already
+                    // ran it there, and re-importing an already-imported module is a no-op.
+                    // Deliberately not calling execusercustomize() — its user-site counterpart,
+                    // consistent with suppressing user-site itself.
+                    site.execsitecustomize();
```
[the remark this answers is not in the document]
alexandru-petre: Traced this one through CPython's `site.py` rather than taking it at face value — **the mechanism is correct**, and it is the one finding in this review that neither of us had already covered. Leaving the thread open for @viogroza, with two qualifications that should change how urgently it is treated.

**The mechanism holds.** `site.execsitecustomize()` is a plain `import sitecustomize`. For a `--system-site-packages` venv `SetNoSiteFlag()` is not called, so `site.main()` runs during `Py_Initialize` while `sys.prefix` still points at the base install (per the `PythonHome` change on line 84) — it adds base and user site-packages, then calls `execsitecustomize()`, which imports and caches `sitecustomize` in `sys.modules`. By the time `PostInitializationVenvSetup` puts the venv's site-packages at `sys.path[0]` and calls `execsitecustomize()` again, the module is already cached and the call is a no-op. Native venv startup orders venv site-packages first, so it picks up the venv's copy. That is a genuine divergence.

**Qualification 1 — it is not a regression.** Before this PR `PostInitializationVenvSetup` did not call `execsitecustomize()` at all, so the venv-local `sitecustomize.py` never ran in this mode either. The new call is a real improvement for the default-venv case (which `Venv_SiteCustomize_Still_Runs` covers); it simply does not help in the `--system-site-packages` case. Nothing that worked before stops working.

**Qualification 2 — the trigger is narrow.** A *failed* import is not cached in `sys.modules`, so when no base-or-user `sitecustomize.py` exists, the second call does find the venv's. The divergence needs a `sitecustomize.py` in base-or-user site **and** one in the venv, on a `--system-site-packages` venv specifically.

**On the proposed fix.** "Defer site processing for this mode too, then add venv/user/base sites in native order" means calling `SetNoSiteFlag()` unconditionally and reimplementing `site.main()`'s ordering by hand, including user-site handling. That is a large blast radius for a patch release fixing a Major customer regression, and it would put the `--system-site-packages` path — currently the one path that still behaves exactly as it did before — onto brand-new code. I would not do that here.

The tempting shortcut of `sys.modules.pop('sitecustomize', None)` before the second call is worse, not better: when the venv has no `sitecustomize.py`, it re-imports the base one and runs its side effects a second time.

**Proportionate response:** amend the comment at lines 377-385. It currently says the call is "safe … a no-op", which is true but frames the no-op as harmless when it actually conceals a precedence divergence. State the limitation explicitly, and add the missing coverage — `Venv_SiteCustomize_Still_Runs` only exercises the default-venv path, so a `--system-site-packages` variant with a `sitecustomize.py` in *both* locations would pin the current behaviour and document it as known. If full native parity is wanted, it deserves its own ticket rather than riding along on this fix.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 359f5fc: added `Engine.HasStdlib` — falls back to `Path.GetDirectoryName(LibraryPath)` (mandatory, validated) whenever the resolved `home` doesn't actually carry a stdlib (`Lib\encodings` on Windows, `lib/python*/encodings` on POSIX). Covers the Microsoft Store case, and the missing-`home`-key case flagged separately below, with the same check.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
viogroza: Fixed in 359f5fc — `TryReadVenvConfig` now falls back to `version_info` when `version` is absent. Added `VenvDetectionTests.VersionInfoKey_Is_Read_As_Fallback_For_Version` and `VenvVersionValidationTests.VersionInfoOnly_CrossChecksJustLikeVersion`, both verified to fail without the fix.

Activities/Python/UiPath.Python/Impl/OutOfProcessEngine.cs
viogroza: Went with option A per your trace — removed `ClearUserSiteEnvironmentOverride` entirely (359f5fc): `Controller.cs` no longer touches `PYTHONNOUSERSITE` at all, so an ambient value now flows to the host untouched, same as a natively-activated `--system-site-packages` venv would behave. Flipped `Venv_With_SystemSitePackages_Ignores_Ambient_PYTHONNOUSERSITE` to `Venv_With_SystemSitePackages_Honors_Ambient_PYTHONNOUSERSITE`, now asserting the `usersite` marker is absent when the var is set — verified failing against the old (stripping) behavior, passing now.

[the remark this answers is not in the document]
viogroza: Same fix, done in 359f5fc — see https://github.com/UiPath/Community.Activities/pull/596#discussion_r3852970950 for the detail.

Activities/Python/UiPath.Python.Tests/VenvVersionValidationTests.cs
viogroza: Fixed in 359f5fc — both now assert via `Assert.Null(Record.Exception(() => EngineProvider.ValidateInstallation(...)))`, matching the pattern you pointed at.

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
viogroza: Not closed — no real installer-based Python 3.10 available in this environment, and I don't want to claim coverage I can't actually run. What I did add: `EngineStdlibDetectionTests` unit-tests `HasStdlib`/`WindowsStdlibLandmark`/`PosixStdlibDirectory` directly against synthetic fixtures, including the exact Microsoft-Store-shaped case (a `home` folder with no `Lib` at all) — each verified to fail without the corresponding fix. That closes the logic gap behind the HIGH finding, but a live run against the ticket's repro package on a real installer-based Python remains open. Flagging as a residual gap rather than claiming it's covered.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
viogroza: Fixed in 359f5fc — wrapped the read in try/catch (`IOException`/`UnauthorizedAccessException`), returning null, matching `ValidatePythonLibrary`/`ValidatePythonExecutable`'s convention right next to it. Added `VenvDetectionTests.UnreadablePyvenvCfg_Is_Not_Detected_And_Does_Not_Throw`, which opens the file with `FileShare.None` to force the failure — verified to throw without the fix, returns null with it.

Activities/Python/UiPath.Python/EngineProvider.cs
viogroza: Added a `## Behaviour changes` section to the PR description naming the new `NotSupportedException` explicitly, per your suggestion — no code change needed since the stricter behavior is intended.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 359f5fc — extracted `ConfigureRuntime(VenvDetection.VenvInfo venv)` covering `SetDllDirectory`/`Runtime.PythonDLL`/`PythonHome`/`SetNoSiteFlag`, with the explanatory prose moved onto its XML doc as you suggested. `Initialize` now reads as detect → configure → initialize → post-setup.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 359f5fc as part of the `ConfigureRuntime` extraction — restated accurately: EngineProvider cross-checks the version declared in `pyvenv.cfg` against `LibraryPath`; it doesn't itself validate the install at `home` — that's what the new `HasStdlib` fallback (see the HIGH finding above) covers instead.

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
viogroza: Fixed in 359f5fc — `EmbeddedPythonRuntimeBootstrap.PythonVersion` is now `public`, plus a new `UserSiteVersionFolder` property deriving the `PythonXY` folder name from it. Both test files reference these instead of restating the version as a separate literal.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 359f5fc as a side effect of the `HasStdlib` fallback above — a `pyvenv.cfg` with `version` but no `home` now yields a null/empty `pythonHome`, `HasStdlib` correctly returns false for it, and the fallback to `LibraryPath`'s own directory kicks in instead of silently skipping the assignment.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Not code-fixed — nothing to fix today, it's correct for every currently-supported version. The risk is now called out explicitly in `ConfigureRuntime`'s XML doc instead: notes the 3.12 deprecation/3.15 removal plan and that extending `_supportedRuntimeVersions` to 3.15+ needs to re-check this flag first, plus the `PyConfig.site_import` replacement path.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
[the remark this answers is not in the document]
viogroza: Same fix as https://github.com/UiPath/Community.Activities/pull/596#discussion_r3852970944, done in 359f5fc.

Activities/Python/UiPath.Python/Impl/Engine.cs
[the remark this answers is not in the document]
viogroza: You're right, and I verified it live rather than taking the claim at face value: with the unconditional pop, the no-own-copy case really does run the base/user `sitecustomize.py` twice — temporarily restored the old, unconditional version and the new regression test failed with 'collection contained 2 items' where it expects 1.

Fixed in 49dde87: gated the pop+reimport on the venv's own site-packages actually containing a `sitecustomize.py`/`usercustomize.py` (`File.Exists` check). When it does, the venv's copy — now first on `sys.path` — wins as intended. When it doesn't, the module `site.main()` already cached (already the correct, highest-priority one in that case) is left untouched, so it runs exactly once.

Added `Venv_With_SystemSitePackages_And_No_Own_SiteCustomize_Does_Not_Rerun_UserSite_Copy` (venv has no `sitecustomize.py`, user site does — asserts it runs exactly once), verified failing against the previous unconditional version and passing against the gated one. Went with the actual fix rather than documenting it as a known limitation, since it turned out to be a small, safe, fully-tested change once I saw the shape of it — but happy to revisit if you'd rather it stayed a documented limitation instead.

Activities/Python/UiPath.Python/Impl/VenvDetection.cs
viogroza: This was actually already fixed in 359f5fc, pushed shortly before this comment landed — `GetVenvInfo` now derives the parent from the trimmed path (`Path.GetDirectoryName(trimmedPath)` instead of `Path.GetDirectoryName(path)`), exactly as you and Copilot both suggested. Added `VenvDetectionTests.LauncherSubfolder_With_TrailingSeparator_Is_Detected` (both `Scripts` and `bin`, with a trailing separator appended), verified to fail without the fix and pass with it.

viogroza:
## Review comments addressed

Two follow-up commits: `359f5fc` (first pass) and `49dde87` (a regression @alexandru-petre caught in the first pass's sitecustomize fix — see below).

**Fixed:**
- `PythonHome` now falls back to `LibraryPath`'s directory when the venv's declared home doesn't carry a real stdlib (Microsoft Store Python, or a `pyvenv.cfg` missing `home` entirely) — HIGH + one of the LOW findings.
- `version_info` read as a fallback for `version` (virtualenv/uv venvs) — both the venv-detection and version-cross-check sides.
- `pyvenv.cfg` read errors (ACLs, AV, concurrent `pip`) now return `null` instead of throwing out of the public `ValidateInstallation`.
- Trailing-separator bug in launcher-folder detection (`Scripts\`/`bin\`) — fixed and covered.
- `ClearUserSiteEnvironmentOverride` removed entirely — an ambient `PYTHONNOUSERSITE` is now honored for `--system-site-packages` venvs exactly like native CPython, no override.
- `sitecustomize.py`/`usercustomize.py` precedence for `--system-site-packages` venvs, **and** a real double-execution regression in that same fix that @alexandru-petre caught afterward (`49dde87`) — the pop+reimport is now gated on the venv actually having its own copy, so a base/user copy's side effects never run twice.
- `Engine.Initialize` extracted into `ConfigureRuntime` (Sonar S3776, the one genuinely PR-caused finding behind the failing quality gate — the rest is confirmed pre-existing debt in touched files, left alone).
- The misleading `EngineProvider`-validated-`home` comment, corrected.
- Hardcoded embedded-runtime version duplicated across two test files, now derived from one source.
- `PostInitializationVenvSetup` made `static`, `GC.SuppressFinalize` added to three tests' `Dispose()` — the remaining genuinely-new Sonar findings.
- PR description updated with a `## Behaviour changes` section for the new `NotSupportedException`.

**Not fixed, called out explicitly rather than silently skipped:**
- No live run against a real installer-based Python 3.10 + genuinely mismatched pywin32 builds — no such environment available here. Added direct, synthetic-fixture unit tests for the `HasStdlib` logic instead (including the Microsoft-Store shape), each verified to fail without its fix — but that's not the same as a live repro.
- `Py_NoSiteFlag`'s CPython 3.12 deprecation/3.15 removal — documented as a checklist item in `ConfigureRuntime`'s XML doc rather than acted on, since nothing needs to change until 3.15+ support is actually added.

Every fix in both commits was verified the same way: temporarily revert it, confirm the corresponding test fails, restore it, confirm green. Full suite: 98/98 passing.

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
viogroza: I tested locally with the scenario mentioned (the original scenario from the customer) and the issue is fixed

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -296,29 +410,59 @@ private static string GetEnvSitePackagesPath(string venvPath)
...
+                    // inserted above, ever got a chance to take precedence. Only when the venv
+                    // itself actually carries its own copy do we drop the cached module and
+                    // re-import, so the venv's copy — now first on sys.path — wins, matching a
+                    // natively-activated venv. Gating on the venv actually having its own copy
+                    // matters: unconditionally popping and re-importing would, when the venv has
+                    // no copy of its own, still find the same base/user module again and run its
+                    // side effects a *second* time. When the venv has no copy, the module already
+                    // cached by site.main() (if any) is already the correct, highest-priority one
+                    // — left untouched. No-op either way for the default venv case: SetNoSiteFlag
+                    // prevented anything from being imported yet, so the cache is empty going in.
```
alexandru-petre: **[Correctness/MEDIUM] The new gate over-applies to the default-venv case, narrowing behaviour on the ticket's primary path**

The `File.Exists` gate exists to stop double-execution in the `--system-site-packages` case, where `site.main()` has already cached a module. But it now gates the **default venv** path too — where `SetNoSiteFlag()` means `site.main()` never ran and nothing is cached, so there was nothing to guard against.

Consequence: for a default venv with no venv-local `sitecustomize.py`, `execsitecustomize()` is no longer called at all. Under `Py_NoSiteFlag` the stdlib paths are still on `sys.path`, so a `<base>\Lib\sitecustomize.py` — corporate images do ship these, and it is exactly the "proxies, logging, etc." case the comment above cites as the reason to preserve this — was imported by the unconditional call in `3b2a9d61` and is now skipped. Same for anything reachable via `PYTHONPATH`.

`Venv_SiteCustomize_Still_Runs` does not catch it, because it writes `sitecustomize.py` into the venv's own site-packages.

**Suggested fix** — split the two cases, which also makes the intent read directly off the code:

```csharp
if (venv.ShouldDisableUserSite)
{
    // site.main() never ran, so nothing is cached: unconditional import is safe here,
    // and matches what a natively-activated default venv does.
    site.execsitecustomize();
}
else if (File.Exists(Path.Combine(sitePackagesPath, "sitecustomize.py")))
{
    // site.main() already cached a base/user copy; drop it so the venv's own copy,
    // now first on sys.path, wins.
    sys.modules.pop("sitecustomize", null);
    site.execsitecustomize();
}
```

Worth a test for the uncovered shape: default venv, **no** venv-local `sitecustomize.py`, one in the base install's `Lib` — assert it still runs.

**Secondary, LOW:** popping from `sys.modules` does not undo side effects the base/user copy has *already executed*. In the `--system-site-packages` branch both copies therefore run, where native runs only the venv's. Closer to parity than before and probably acceptable, but the comment presents the pop as full parity — one sentence acknowledging "both ran" would keep it honest.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -296,29 +410,59 @@ private static string GetEnvSitePackagesPath(string venvPath)
...
+                    if (File.Exists(Path.Combine(sitePackagesPath, "sitecustomize.py")))
+                    {
+                        sys.modules.pop("sitecustomize", null);
+                        site.execsitecustomize();
+                    }
+
+                    // execusercustomize() mirrors site.main()'s own "if ENABLE_USER_SITE:" guard —
```
alexandru-petre: **[Maintainability/MEDIUM] The `usercustomize` block checks the wrong directory, and its guard is wrong in the one case it could fire**

Two independent problems converge here — the second is Copilot's suppressed comment on line 461, which I verified and am folding in.

**1. Wrong directory (effectively dead code).** `sitePackagesPath` is the **venv's** `Lib\site-packages`, but `usercustomize.py` lives in the *user-site* directory by definition (`%APPDATA%\Roaming\Python\PythonXY\site-packages`). The `File.Exists` check is therefore essentially never true, so the block never runs. And in the only mode it targets, `site.main()` has already called `execusercustomize()` — so even firing correctly would be redundant.

**2. The guard doesn't mirror what it claims.** The comment says it mirrors `site.main()`'s `if ENABLE_USER_SITE:` check, but `!venv.ShouldDisableUserSite` is a static approximation of a runtime value. With a `--system-site-packages` venv and ambient `PYTHONNOUSERSITE=1`, `site.ENABLE_USER_SITE` is `False` — yet this guard still passes. So in the rare case the block does fire, it fires precisely when user-site customization should be suppressed. That directly undercuts the ambient-`PYTHONNOUSERSITE` behaviour this PR just went to some trouble to honour.

It also silently reverses `3b2a9d61`'s explicit *"Deliberately not calling execusercustomize() — its user-site counterpart, consistent with suppressing user-site itself"* without saying why, and nothing tests it.

**Suggested fix:** delete the block and restore the original one-line rationale. Given (1) it is unreachable in practice and (2) reachable only when it would be wrong, fixing the guard is not worth it — removal is the smaller and safer change. If you do want it, the guard has to read the runtime value (`(bool)site.ENABLE_USER_SITE`) rather than the config flag, and check the user-site directory rather than the venv's.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -96,6 +99,77 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
+                Runtime.PythonDLL = _libraryPath;
+
+            var pythonHome = venv != null ? ResolvePythonHome(venv.Home) : _path;
+            if (!HasStdlib(pythonHome))
+                pythonHome = Path.GetDirectoryName(_libraryPath);
+
+            if (!pythonHome.IsNullOrEmpty())
+                PythonEngine.PythonHome = pythonHome;
+
+            if (venv != null && venv.ShouldDisableUserSite)
```
alexandru-petre, line 170: **[Correctness/MEDIUM] `SetNoSiteFlag` also removes the `quit`/`exit`/`help` builtins from default-venv scopes**

Surfacing this from Copilot's **suppressed** section (`Engine.cs:170`) — it is collapsed in the review UI, and of its three suppressed comments this is the one with real user-facing impact. I verified the mechanism.

`site.main()` does more than path setup. It also calls `setquit()`, `setcopyright()`, `sethelper()` and `enablerlcompleter()`. `Py_NoSiteFlag` skips all of it, and `PostInitializationVenvSetup` restores only the path handling and the customization imports — so `quit`, `exit`, `help`, `copyright`, `credits` and `license` are simply absent from a default-venv scope.

**Why it matters:** a user script calling `exit()` — not good practice, but common in RPA scripts — now dies with `NameError: name 'exit' is not defined`. It works today, it works natively in a venv, and it works in every non-venv scope. This is a silent behaviour change introduced by this PR, on exactly the default-venv path the ticket targets, and it produces an error that gives the user no hint about its cause.

**Suggested fix:** restore the non-path portions in `PostInitializationVenvSetup` for the `ShouldDisableUserSite` branch — they are plain module-level functions in `site.py` and cheap to call:

```csharp
site.setquit();
site.setcopyright();
site.sethelper();
```

`enablerlcompleter()` only registers an interactive-startup hook, so it can be skipped. Worth a test asserting `exit` and `help` resolve inside a default-venv scope.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -296,29 +410,59 @@ private static string GetEnvSitePackagesPath(string venvPath)
...
+            if (venv != null)
             {
                 using (Py.GIL())
                 {
                     dynamic sys = Py.Import("sys");
                     dynamic site = Py.Import("site");
 
-                    sys.prefix = venvPath;
-                    sys.exec_prefix = venvPath;
+                    sys.prefix = venv.Root;
```
alexandru-petre, line 464: **[Correctness/LOW] Assigning `sys.prefix` leaves `site.PREFIXES` pointing at the base install**

Also from Copilot's **suppressed** section (`Engine.cs:423`); verified and agreed, at low severity.

`site.PREFIXES` is initialised at module-import time from `[sys.prefix, sys.exec_prefix]`. `Py.Import("site")` on the line above runs while `PythonHome` still points at the base install, and the `sys.prefix` assignment here does not retroactively update it. So `site.getsitepackages()` keeps reporting base-install locations and omits the venv — for default venvs too, not just `--system-site-packages`.

This does not affect script execution or the STUD=81085 fix, which is why LOW: it is a fidelity gap that shows up only for code introspecting the environment (packaging helpers, `pip`-adjacent tooling, diagnostics that print `getsitepackages()`).

**Suggested fix** — one line, right after the `sys.exec_prefix` assignment:

```csharp
// site.PREFIXES was captured at import time, while PythonHome still pointed at the base
// install; resync it so site.getsitepackages() reports the venv like native activation does.
site.PREFIXES = venv.IncludeSystemSitePackages
    ? new[] { venv.Root, basePrefix }
    : new[] { venv.Root };
```

Reasonable to defer to a follow-up if you would rather keep this patch tight — but it is cheap, and it is the kind of thing that is much harder to justify revisiting later.

alexandru-petre (commented):
## Review of `359f5fc3` + `49dde878`

Thanks — every finding from the first round is either fixed or consciously deferred with a stated reason, and the deferrals are the right calls. Re-reviewed the two new commits and also validated Copilot's latest review, **including its three suppressed comments**, since those are collapsed and easy to miss.

### Verified correct

- `Controller.cs` reverted to the exact pre-PR blob (`a1a7d1a9` → `e7510269`) — no residue. `OutOfProcessEngine` also drops its now-redundant `GetVenvInfo` call, removing a duplicate `pyvenv.cfg` read in the parent.
- The `HasStdlib` fallback goes further than asked, in a good way: since both call sites pass `Version.Auto`, `_path` always contains a Python exe, so the only non-venv case where the fallback fires is one that was already broken (e.g. `Path` pointed at a Store alias folder). Strict improvement on Windows.
- The `try/catch` wraps the `foreach`, not just the `File.ReadLines()` call — correct placement for a lazy enumerator, and the usual place people get it wrong.
- Tests are strong: `UnreadablePyvenvCfg_...` takes a real `FileShare.None` lock rather than mocking; `VersionInfoOnly_CrossChecksJustLikeVersion` asserts the *throw*, proving the cross-check engages rather than merely parsing; `Venv_With_SystemSitePackages_And_No_Own_SiteCustomize_Does_Not_Rerun_UserSite_Copy` uses `Assert.Single` to pin "exactly once", which is precisely the double-execution regression. The PYTHONNOUSERSITE test was correctly inverted **and** renamed to `..._Honors_...`. The `WindowsStdlibLandmark`/`PosixStdlibDirectory` casing tests, justified by NTFS case-folding on Windows-only CI, are a genuinely thoughtful bit of design.

### New findings (4 inline below)

All four live in the `49dde878` sitecustomize work or in what `SetNoSiteFlag` silently removes. Two are mine, two are Copilot suppressed comments I verified and am surfacing.

### Copilot's latest review — validated

| Copilot finding | Verdict |
|---|---|
| POSIX `Path.GetDirectoryName(_libraryPath)` is not a valid prefix (`Engine.cs:164`) | **Valid** — endorsed in-thread with a minimal fix |
| Tests inherit ambient `PYTHONNOUSERSITE` (`VenvUserSiteIsolationTests.cs:45`) | **Valid** — endorsed in-thread; now load-bearing since the fix deliberately honours the ambient value |
| *Suppressed:* `SetNoSiteFlag` also drops `quit`/`exit`/`help` builtins (`Engine.cs:170`) | **Valid** — surfaced below, the strongest of the three |
| *Suppressed:* `sys.prefix` assignment leaves `site.PREFIXES` stale (`Engine.cs:423`) | **Valid** — surfaced below as LOW |
| *Suppressed:* `usercustomize` guard doesn't mirror `ENABLE_USER_SITE` (`Engine.cs:461`) | **Valid** — merged into my finding on the same block |

One recurring defect in Copilot's output worth knowing about: its *"This issue also appears in the following locations"* lists are unreliable. On `Engine.cs:164` it points at lines 169, 422 and 458 — which are its own three *suppressed* findings, not other instances of the POSIX problem. It did the same thing last round. Read those cross-references with suspicion.

Activities/Python/UiPath.Python/Impl/Engine.cs
```diff
@@ -96,6 +99,77 @@ public async Task Initialize(string workingFolder, CancellationToken ct, double
...
+        {
+            if (_isWindows && !_path.IsNullOrEmpty())
+                SetDllDirectory(Path.GetFullPath(_path));
+
+            if (!_libraryPath.IsNullOrEmpty())
+                Runtime.PythonDLL = _libraryPath;
+
+            var pythonHome = venv != null ? ResolvePythonHome(venv.Home) : _path;
+            if (!HasStdlib(pythonHome))
+                pythonHome = Path.GetDirectoryName(_libraryPath);
```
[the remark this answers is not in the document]
alexandru-petre: Verified — this is correct, and it is the one gap in the `HasStdlib` fix. Endorsing rather than re-filing.

On Windows the fallback is right (`C:\Python310\python310.dll` → `C:\Python310`). On POSIX it is shaped wrong: `/usr/lib/x86_64-linux-gnu/libpython3.12.so` → `/usr/lib/x86_64-linux-gnu`, where the prefix is `/usr`. And because nothing re-validates after the assignment, a bad `home` on Linux/macOS just swaps one non-stdlib `PythonHome` for another.

Worth noting how narrow the trigger is, so this gets sized correctly: `ResolvePythonHome` already strips a trailing `bin`, so a normal POSIX venv yields `/usr`, `HasStdlib` passes, and the fallback never fires. It only fires when `home` is missing or bogus — and the motivating case (a Microsoft Store app-execution-alias folder) is Windows-only.

**Minimal fix** — validate the fallback with the same predicate instead of trusting it, so a known-bad value is never assigned:

```csharp
var pythonHome = venv != null ? ResolvePythonHome(venv.Home) : _path;
if (!HasStdlib(pythonHome))
{
    var fromLibrary = Path.GetDirectoryName(_libraryPath);
    pythonHome = HasStdlib(fromLibrary) ? fromLibrary : null;   // leave unset rather than set a known-bad prefix
}
```

That keeps the Windows win, and on POSIX falls back to the pre-existing "leave `PythonHome` unset and let CPython compute the prefix from the DLL" behaviour — which is better than pointing it somewhere definitively wrong. Walking ancestors for `lib/python*/encodings`, as suggested here, would also work and covers more; the guard above is the smaller change and removes the "silently wrong" property either way.

One thing to disregard in this comment: the *"also appears in the following locations — line 169, line 422, line 458"* list is not other instances of this problem. Those three lines are this review's own suppressed comments (`SetNoSiteFlag`, `sys.prefix`, the `usercustomize` guard), which are unrelated to POSIX prefix resolution. I have surfaced them separately on their own lines.

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
```diff
@@ -0,0 +1,389 @@
...
+
+        public VenvUserSiteIsolationTests()
+        {
+            _rootDir = Directory.CreateDirectory(Path.Combine(Path.GetTempPath(), "venv-usersite-tests", Guid.NewGuid().ToString("N"))).FullName;
+            _venvDir = Path.Combine(_rootDir, ".venv");
+            _userBaseDir = Path.Combine(_rootDir, "userbase");
+            _markerFile = Path.Combine(_rootDir, "marker.txt");
+
+            _previousUserBase = Environment.GetEnvironmentVariable("PYTHONUSERBASE");
+            _previousNoUserSite = Environment.GetEnvironmentVariable("PYTHONNOUSERSITE");
```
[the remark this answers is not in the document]
alexandru-petre: Verified — agreed, and this matters more now than it would have a commit ago. Endorsing rather than re-filing.

The constructor saves `PYTHONUSERBASE` and `PYTHONNOUSERSITE` but never clears them, so every test in this class inherits whatever the agent or dev machine happens to have set. Two concrete failure modes:

- With ambient `PYTHONNOUSERSITE=1`, `Venv_With_SystemSitePackages_Still_Processes_UserSite_PthFiles` and `Venv_With_SystemSitePackages_OwnSiteCustomize_Still_Runs` **fail**, because they assert user-site processing happens.
- `Venv_Does_Not_Process_UserSite_PthFiles` **passes vacuously** — its `Assert.DoesNotContain("usersite", …)` succeeds for the wrong reason, so the isolation it exists to prove goes unverified.

The second is the worse one: a green suite that is not actually testing the fix.

This became load-bearing in `359f5fc3`. Before that, the code stripped `PYTHONNOUSERSITE` from the child's environment, so an ambient value could not reach the interpreter. Now that we deliberately let it flow through — correctly, per the `site.py` semantics — the ambient value genuinely changes outcomes, and the tests have to control it.

**Fix**, in the constructor right after the two `GetEnvironmentVariable` calls (`Dispose` already restores both, so nothing else changes):

```csharp
// Both are read by the spawned interpreter and would otherwise leak in from the agent,
// silently inverting the user-site expectations below.
Environment.SetEnvironmentVariable("PYTHONUSERBASE", null);
Environment.SetEnvironmentVariable("PYTHONNOUSERSITE", null);
```

`Venv_With_SystemSitePackages_Honors_Ambient_PYTHONNOUSERSITE` sets the variable itself, so it is unaffected by the clearing.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 0a1e7ba. Confirmed you're right — verified live by temporarily restoring the merged gate and watching a new test fail. Split into two branches: default venv (`ShouldDisableUserSite`) now calls `execsitecustomize()` unconditionally again, matching `3b2a9d61`'s original behavior and native venv activation; `--system-site-packages` keeps the gated pop+reimport, since that's the only case where anything could actually be cached to guard against. Added `Venv_SiteCustomize_Still_Runs_From_BaseInstall_When_Venv_Has_None` (base install's `Lib\sitecustomize.py`, no venv-local copy) — verified it fails against the merged gate and passes against the split version.

On the secondary LOW point: agreed, and now moot for the default-venv case since that branch never pops/reimports at all. For `--system-site-packages`, the comment now says plainly "the module already cached... is left untouched, avoiding running its side effects a second time" — i.e. only the venv branch pops, and only when the venv has its own copy, so there's no case left where both run.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 0a1e7ba — removed the block entirely and restored the original one-line rationale, per your "removal is the smaller and safer change" suggestion. Confirmed both your points: `usercustomize.py` only ever lives in the user-site directory (never a venv's own site-packages), so the check as written was dead code; and even on the rare path it could fire, the guard was a static config flag rather than the runtime `ENABLE_USER_SITE` value, so it could fire exactly backwards relative to ambient `PYTHONNOUSERSITE`. Nothing left for venv-specific setup to do here in either case — site.main() (when it ran) already handled it via its own guard.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Fixed in 0a1e7ba — added `site.setquit()`/`site.setcopyright()`/`site.sethelper()` to the default-venv branch, right before `execsitecustomize()` to match `site.main()`'s own ordering. Left `enablerlcompleter()` out, per your note that it only matters interactively. Added `Venv_ExitAndHelp_Builtins_Still_Resolve` (asserts `hasattr(builtins, 'exit')`/`'help'` inside a default-venv scope), verified failing without the fix and passing with it.

Activities/Python/UiPath.Python/Impl/Engine.cs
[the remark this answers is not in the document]
viogroza: Confirmed, and fixed in 0a1e7ba (superseding the simpler validate-after-fallback I'd have reached for otherwise) — went with the full ancestor walk since it covers more without much extra cost: `ResolvePrefixFromLibraryPath` now walks up from the library's directory, testing `HasStdlib` at each level, instead of trusting the immediate parent. On Windows this resolves on the first iteration as before (no behavior change); on POSIX it correctly climbs past `lib`/multiarch triplets to the real prefix. Also closed a second gap this exposed once actually tested: the embeddable distribution's own `._pth`-based home has no literal `Lib\encodings` either, so `HasStdlib` now also recognizes a `*._pth` file as valid evidence of a home. Added direct tests for both the ancestor-walk success/exhaustion cases and the `._pth` recognition, each verified against a reverted version first.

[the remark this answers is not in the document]
viogroza: Same fix, done in 0a1e7ba — see https://github.com/UiPath/Community.Activities/pull/596#discussion_r3854699875 for the detail. Went with the full ancestor walk rather than the minimal validate-and-null-out version, since it covers more (multiarch triplets etc.) for about the same amount of code.

Activities/Python/UiPath.Python.Tests/VenvUserSiteIsolationTests.cs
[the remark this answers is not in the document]
viogroza: Fixed in 0a1e7ba — the constructor now clears both `PYTHONUSERBASE` and `PYTHONNOUSERSITE` right after saving them, so every test starts from a known-clean baseline regardless of the host machine's ambient environment. `Dispose` still restores whatever was actually ambient; `Venv_With_SystemSitePackages_Honors_Ambient_PYTHONNOUSERSITE` sets the variable itself and is unaffected.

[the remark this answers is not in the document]
viogroza: Same fix, done in 0a1e7ba — see https://github.com/UiPath/Community.Activities/pull/596#discussion_r3854699955 for the detail. Agreed this is the worse of the two failure modes you called out — a vacuously-passing isolation test is worse than a flaky-failing one, since nothing flags that it stopped proving anything.

Activities/Python/UiPath.Python/Impl/Engine.cs
viogroza: Deferring this one rather than folding it in now, per your own "reasonable to defer to a follow-up" note. Implementing it cleanly needs the resolved base prefix threaded into `PostInitializationVenvSetup` (it currently only receives `venv`), which is a bigger plumbing change than the snippet implies for a LOW-severity, introspection-only gap — doesn't affect script execution or the STUD-81085 fix itself. Given how much ground the last two rounds already covered (including two rounds of my own fixes needing a fix), I'd rather not rush a fourth change into this same method without it getting the same scrutiny. Happy to open a follow-up ticket for it, or fold it in here if you'd rather have it now — your call.

viogroza:
## Second round of review comments addressed

Commit `0a1e7ba`, on top of the previous `359f5fc`/`49dde87`.

**Fixed — new findings from this round:**
- My own `File.Exists` gate (from `49dde87`) over-applied to the default-venv case, silently dropping `execsitecustomize()`'s unconditional call for a venv with no local `sitecustomize.py` but a base-install `Lib\sitecustomize.py` (corporate proxy/logging setups). Split into two branches by `venv.ShouldDisableUserSite`.
- The `usercustomize` block checked the venv's own site-packages — `usercustomize.py` only ever lives in the per-user site directory, so it was dead code, and wrong on the rare path it could fire. Removed entirely, restored the original one-line rationale.
- `SetNoSiteFlag` silently drops `exit`/`quit`/`help`/`copyright`/`credits`/`license` builtins from a default-venv scope (Copilot's suppressed finding, verified and surfaced by @alexandru-petre). Restored `setquit()`/`setcopyright()`/`sethelper()`.
- `ResolvePrefixFromLibraryPath`'s one-level fallback was wrong on POSIX (library sits below the real prefix, not at it) — now walks up ancestors testing `HasStdlib`. Also closed a gap this exposed once tested: `HasStdlib` didn't recognize a legitimate `._pth`-based home (the embeddable distribution's own layout), now fixed.
- `VenvUserSiteIsolationTests`'s constructor now clears `PYTHONUSERBASE`/`PYTHONNOUSERSITE` after saving them, so tests don't inherit whatever the host machine's ambient environment has.

**Deferred, with the reviewer's own suggestion to do so:**
- `site.PREFIXES` staying stale after the `sys.prefix` reassignment (LOW, introspection-only — doesn't affect script execution or the fix itself). Needs threading the resolved base prefix into `PostInitializationVenvSetup`, which didn't seem worth rushing in on top of everything else this round.

Every fix verified the same way as before: temporarily revert, confirm the corresponding test fails, restore, confirm green. Full suite: 104/104 passing.

alexandru-petre (approved):
## Approving ✅

Four rounds, 24 threads, 21 resolved. Every MEDIUM and the one HIGH are fixed and verified against `0a1e7ba1` — not just marked done. What stands out is that the fixes were checked rather than assumed: the merged sitecustomize gate was confirmed by restoring it and watching a new test fail, and the double-execution regression in `49dde878` was caught and corrected in the same session it was introduced.

The `._pth` recognition in `HasStdlib` deserves a specific mention — it was the author's own catch, and it is the sharpest thing in the PR. Without it, commits 2–3 were passing their venv tests through the *fallback* path, since the embeddable runtime has no unpacked `Lib`. The primary `home`-based resolution was silently untested and nobody had flagged it.

### Nits — all non-blocking, all one-to-three lines

None of these justify another round on their own; batch them into a follow-up if you prefer.

**1. `[Correctness/LOW]` Stdlib landmark is satisfied by an empty directory** — `Engine.cs:382`, same shape at `:401`

`HasStdlib` accepts any folder containing `Lib\encodings` (POSIX: `lib/python*/encodings`) regardless of contents, so an empty directory of that name passes and `PythonHome` gets set to a home that still cannot bootstrap. `Venv_SiteCustomize_Still_Runs_From_BaseInstall_When_Venv_Has_None` shows how cheap that is to satisfy — it creates `fakebase\Lib\encodings` empty purely to get past the check. Testing for `Lib\encodings\__init__.py` instead would tighten a check the whole fallback now hinges on.

**2. `[Bugs/LOW]` The ancestor walk aborts on one unreadable ancestor** — `Engine.cs:378`, reached from `:417`

`ResolvePrefixFromLibraryPath` calls `HasStdlib` at each level, and `HasStdlib` now does `Directory.EnumerateFiles(pythonHome, "*._pth")`, which throws `UnauthorizedAccessException` on a directory the robot account cannot enumerate. One denied ancestor aborts the whole walk instead of continuing upward, failing init even when a higher ancestor would have resolved. Before `0a1e7ba1` the fallback did no I/O at all. A `try/catch { continue; }` around the per-level probe matches what `TryReadVenvConfig` and the `EngineProvider` validators already do.

**3–4. The two open Copilot findings** — both valid, both LOW, both still unaddressed:

- [Bare relative `LibraryPath`](https://github.com/UiPath/Community.Activities/pull/596#discussion_r3855169681) (`Engine.cs:422`). One correction to that comment: `Path.GetDirectoryName("python314.dll")` returns `""`, not `null` — but the effect is exactly as described, the `while` guard fails immediately and the fallback silently no-ops in the case it exists to repair. `Path.GetFullPath` first.
- [`sitecustomize` as a package or bytecode module](https://github.com/UiPath/Community.Activities/pull/596#discussion_r3855169726) (`Engine.cs:498`). Narrow — needs `--system-site-packages` **and** a non-source form **and** a cached base/user copy.

### Two deferrals that need a home after merge

Both were deferred for good reasons, and I agree with both calls — but once this merges, the threads carrying them stop being somewhere anyone looks:

- [`site.PREFIXES` left stale](https://github.com/UiPath/Community.Activities/pull/596#discussion_r3854783882) — fidelity only, no execution impact; needs the resolved base prefix threaded through to fix cleanly.
- **No regression coverage against an installer-based Python.** This is the one I would still like closed, and it is the reason I am calling it out here rather than letting it disappear. The suite runs entirely against the embeddable distribution, whose `._pth` bootstrap resolves the stdlib independently of `PythonHome` — structurally the opposite of the reported configuration (3.10, installer-based, real pywin32). Declining to claim coverage you cannot run was the right instinct. The ticket's own `Test-Pywin32VenvRepro (2).zip` exists; one manual run recorded in the PR body would close it, and for a customer-facing patch that is worth the twenty minutes.

Suggest a follow-up ticket carrying both, linked from this PR.

### On the quality gate

Still red, still not this PR's doing. Verified against the SonarCloud API at this head: all seven issues driving the D Security / D Reliability ratings are **pre-existing** code in touched files — `S4790` (MD5 in `Engine.Hash`), `S6444` (regex without timeout, ×2), `S2551` (`lock(this)`, ×3), `S2445`. The one genuinely PR-caused issue, `S3776` on `Engine.Initialize`, is `CLOSED / FIXED` after the `ConfigureRuntime` extraction. Merge will need either a separate cleanup PR or a gate exception — please do not absorb those into this one.

PR #597, opened 2026-08-22
chore(l10n): sync from Localization

andrei-stoian:
The build that generated this PR: https://uipath.visualstudio.com/DevOps/_build/results?buildId=13105034

### DIFF
--- a/Activities/FTP/UiPath.FTP.Activities/Properties/UiPath.FTP.Activities.de.resx
+++ b/Activities/FTP/UiPath.FTP.Activities/Properties/UiPath.FTP.Activities.de.resx
@@ -222,6 +222,9 @@
   <data name="NoValidAuthenticationMethod" xml:space="preserve">
     <value>Keine gültige Authentifizierungsmethode gefunden: Sie müssen entweder eine private Schlüsseldatei (und optional eine Passphrase) oder ein Kennwort angeben.</value>
   </data>
+  <data name="AnonymousLoginNotSupportedOnSftp" xml:space="preserve">
+    <value>Anonyme Anmeldung über SFTP wird nicht unterstützt. Deaktivieren Sie die Option „Anonyme Anmeldung verwenden“ oder „SFTP verwenden“.</value>
+  </data>
   <data name="NewPath" xml:space="preserve">
     <value>Neuer Remotepfad</value>
   </data>
@@ -321,10 +324,10 @@
     <value>Allgemein</value>
   </data>
   <data name="Activity_WithFtpSession_Property_DisplayName_Name" xml:space="preserve">
-    <value>Scope der FTP-Sitzung</value>
+    <value>Use FTP connection</value>
   </data>
   <data name="Activity_Delete_Description" xml:space="preserve">
-    <value>Entfernt eine angegebene Datei von einem FTP-Server. Diese Aktivität funktioniert nur, wenn sie innerhalb einer Scope-Aktivität vom Typ „With FTP Session“ platziert wird.</value>
+    <value>Entfernt eine Datei oder einen Ordner von einem FTP-Server. Diese Aktivität funktioniert nur, wenn sie in einer „Use FTP Connection“-Aktivität platziert wird.</value>
     <comment>Activity description</comment>
   </data>
   <data name="Activity_Delete_Name" xml:space="preserve">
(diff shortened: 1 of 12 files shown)


Activities/FTP/UiPath.FTP.Activities/Properties/UiPath.FTP.Activities.ja.resx
```diff
@@ -321,10 +324,10 @@
     <value>共通</value>
   </data>
   <data name="Activity_WithFtpSession_Property_DisplayName_Name" xml:space="preserve">
-    <value>FTP セッション スコープ</value>
+    <value>Use FTP Connection</value>
   </data>
   <data name="Activity_Delete_Description" xml:space="preserve">
-    <value>指定したファイルを FTP サーバーから削除します。このアクティビティは、[FTP セッションを使用] スコープ アクティビティ内に配置した場合にのみ動作します。(Delete)</value>
+    <value>Removes a file or folder from an FTP server. This activity only works if it is placed inside a Use FTP Connection activity.</value>
```
LiviuPonova, line 330: This "sync" reverts a large number of previously-translated Japanese strings back to raw English. `Activity_Delete_Description` here changes from the existing Japanese translation to the literal English `Removes a file or folder from an FTP server. This activity only works if it is placed inside a Use FTP Connection activity.` — the same pattern repeats for ~70+ keys in this file (Activity_DirectoryExists_*, Activity_EnumerateObjects_*, Activity_FileExists_*, Activity_UploadFiles_*, Activity_DownloadFiles_*, Activity_MoveItem_*, most Activity_WithFtpSession_Property_* names/descriptions, AnonymousLoginNotSupportedOnSftp). A ja-locale Studio user would now see the FTP package almost entirely in English where it previously had real Japanese text. Worth confirming with the localization pipeline before merging — this looks like an untranslated/English-fallback export rather than an actual translation sync.

Activities/FTP/UiPath.FTP.Activities/Properties/UiPath.FTP.Activities.de.resx
```diff
@@ -321,10 +324,10 @@
     <value>Allgemein</value>
   </data>
   <data name="Activity_WithFtpSession_Property_DisplayName_Name" xml:space="preserve">
-    <value>Scope der FTP-Sitzung</value>
+    <value>Use FTP connection</value>
```
LiviuPonova, line 327: `Activity_WithFtpSession_Property_DisplayName_Name` — the display name of the renamed 'Use FTP Connection' activity itself — was left as the untranslated English string here instead of getting a real translation (previously this was "Scope der FTP-Sitzung"). The same untranslated leftover shows up in es-MX, fr, ja, ko, pt-BR, pt, and tr, while es, ro, zh-CN, and zh-TW correctly received a translated value (e.g. es: "Usar conexión FTP", ro: "Folosește conexiunea FTP"). Since this is the literal name Studio shows for the activity in the Activities panel/search, German/French/Japanese/Korean/Portuguese/Turkish users would now see an English activity name where they previously saw a localized one.

Activities/FTP/UiPath.FTP.Activities/Properties/UiPath.FTP.Activities.de.resx
```diff
@@ -782,10 +785,79 @@
...
+  <data name="ActivitySynonymMoveFileOrFolder" xml:space="preserve">
+    <value>Element verschieben</value>
+    <comment>Activity search alias</comment>
+  </data>
+  <data name="ActivitySynonymUseFtpConnection" xml:space="preserve">
+    <value>Mit FTP-Verbindung</value>
+    <comment>Activity search alias</comment>
+  </data>
+  <data name="ActivitySynonymCheckIfFolderExists" xml:space="preserve">
+    <value>Directory Exists</value>
```
LiviuPonova, line 852: Of the six new activity-search synonym keys added in this block, three are correctly translated (ActivitySynonymDeleteFileOrFolder='Löschen', ActivitySynonymMoveFileOrFolder='Element verschieben', ActivitySynonymUseFtpConnection='Mit FTP-Verbindung'), but ActivitySynonymCheckIfFolderExists (line 852), ActivitySynonymCheckIfFileExists, and ActivitySynonymListFilesAndFolders are left as raw English ('Directory Exists', 'File Exists', 'Enumerate Objects'). These synonyms exist so a German-locale user typing the activity's old name into Studio's activity search can still find the renamed activity — shipping the English literal instead of a German synonym means search-by-old-name silently doesn't work for these three renamed activities. Same inconsistency (3 of 6 synonym keys left in English) repeats in es-MX, fr, ja, ko, pt-BR, pt, and tr.

LiviuPonova (changes requested):
This localization sync ships a real content regression rather than just refreshing translations — see the three inline comments for specifics:

- The Japanese resx reverts ~70+ previously-translated strings back to raw English (e.g. `Activity_Delete_Description` and most `Activity_WithFtpSession_Property_*` names/descriptions).
- The renamed activity's own display name (`Activity_WithFtpSession_Property_DisplayName_Name`, shown as "Use FTP Connection") is left untranslated in 8 of 12 locales (de, es-MX, fr, ja, ko, pt-BR, pt, tr).
- 3 of the 6 new activity-search synonym keys per locale (CheckIfFolderExists / CheckIfFileExists / ListFilesAndFolders) are left as raw English in 7 locales, while their sibling synonyms in the same new block were correctly translated.

Net effect: German, French, Japanese, Korean, Portuguese, and Turkish Studio users would see the FTP package regress to English in several places, and lose search-by-old-name for three renamed activities. This looks like an English-fallback bug in the localization export rather than an actual translation sync, so requesting changes rather than approving — recommend fixing/re-running the upstream l10n pipeline before merging, or manually restoring the regressed values.

alexandru-petre:
PR replaced by a newer one

PR #600, opened 2026-09-18
Database: Fix NullReferenceException in ExecuteNonQuery with ContinueOnError=True

viogroza:
## Summary
- `ExecuteNonQuery` with `ContinueOnError="True"` threw `System.NullReferenceException` instead of swallowing the SQL error, defeating the purpose of `ContinueOnError`.
- Root cause: on failure, `ConnectionHelper.HandleException` swallows the exception but leaves `affectedRecords` null; the result-building action then unconditionally read `affectedRecords.Result`.
- **Output-contract decision (updated from the initial null-guard approach, per review discussion below):** `affectedRecords` is now initialized to `new DBExecuteCommandResult()` instead of `null`, so on the swallowed-failure path `AffectedRecords` resets to `0` rather than being left untouched/stale. This matches `InsertDataTable`/`BulkInsert`/`BulkUpdate`, which all reset their row-count output to `0` on a swallowed error, and avoids `AffectedRecords` silently keeping a prior run's value (e.g. in a loop, a failed iteration would otherwise still report the previous iteration's row count as if it succeeded). This intentionally diverges from `ExecuteQuery`'s current stale-on-failure behavior for its own outputs — that behavior isn't itself validated as correct, so this PR doesn't propagate it into `ExecuteNonQuery`.
- `Out`/`InOut` entries in `Parameters` are intentionally left out of scope for this reset: `SetOutputParameters` only writes entries present in `ParametersBind`, which stays empty on the failure path either way, so those still remain untouched/stale. There's no natural "reset" default for an arbitrary parameter type the way `0` is for `AffectedRecords`; treated as a separate design question if it needs addressing.

## Test plan
- [x] `ExecuteNonQuery_ContinueOnError_SwallowsExceptionInsteadOfThrowingNullReferenceException`: reproduces the ticket's exact repro (`SELECT * FROM ThisTableDoesNotExist_Probe` with `ContinueOnError=True`) and confirms it no longer throws.
- [x] `ExecuteNonQuery_ContinueOnError_ResetsAffectedRecordsInsteadOfLeavingStaleValue`: binds `AffectedRecords` to a `Variable<int>` pre-seeded with `42` inside a `Sequence`, then asserts it resets to `0` after a swallowed failure — pins down the reset-vs-stale behavior so the assertion can't pass by coincidence of `WorkflowInvoker`'s default `OutArgument` value.
- [x] Full `Activities.Database.sln` test suite passes (51/51).
- [x] Validated end-to-end against a real UiPath project (SQLite-backed workflow): reproduces the `NullReferenceException` against the published GA `2.1.1` package, and resolves cleanly against the fixed build.

Fixes STUD-81504: https://uipath.atlassian.net/browse/STUD-81504


Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
```diff
@@ -84,6 +84,8 @@ protected async override Task<Action<AsyncCodeActivityContext>> ExecuteAsync(Asy
                 }
                 var result = new Action<AsyncCodeActivityContext>(asyncCodeActivityContext =>
                 {
+                    if (affectedRecords == null) return;
```
alexandru-petre, line 87: The guard removes the crash, but it also means that on the swallowed path neither `AffectedRecords.Set(...)` nor `ConnectionHelper.SetOutputParameters(...)` runs. In Workflow Foundation an `OutArgument` writes into a workflow variable only when the activity calls `Set`; if the activity never calls it, the variable is not reset, it keeps the value it already held. So in a loop where iteration 1 updated 5 rows and iteration 2's statement fails with `ContinueOnError = True`, `AffectedRecords` still reads 5, and a downstream `If AffectedRecords > 0` treats the failed statement as a success. The same applies to the `Out` and `InOut` entries in `Parameters`, which stored procedure callers read back after execution.

The other three activities in this pack behave the opposite way: `InsertDataTable`, `BulkInsert` and `BulkUpdate` hold the row count in a plain `int` or `long` local initialised to 0, so after a swallowed error they explicitly write 0 to an identically named and identically typed output. The property description ("Specifies whether the automation should continue even when the activity throws an error") says nothing about output values, so there is no documented contract to appeal to either way. That is exactly why it is worth choosing deliberately rather than inheriting whichever shape was copied.

Both choices are defensible, and mirroring `ExecuteQuery` is what the ticket suggested, so keeping this as it stands needs nothing more than a sentence in the PR description recording that the stale value is intended. If you would rather match the sibling activities, initialising `affectedRecords` to `new DBExecuteCommandResult()` instead of `null` yields `Result == 0` and an empty `ParametersBind`, so both existing calls stay on the happy path and the guard is no longer needed.

Activities/Database/UiPath.Database.Tests/SqliteIntegrationTests.cs
```diff
@@ -320,6 +320,23 @@ public void DatabaseTransaction_InternalConnection_OutputNotBound_IsDisposed()
...
+                ExistingDbConnection = new InArgument<DatabaseConnection>(_ => _fixture.Connection),
+                Sql = new InArgument<string>("SELECT * FROM ThisTableDoesNotExist_Probe"),
+                ContinueOnError = new InArgument<bool>(true),
+                AffectedRecords = new OutArgument<int>()
+            };
+
+            var outputs = WorkflowInvoker.Invoke(activity, TimeSpan.FromSeconds(30));
+
+            var affected = (int)outputs[nameof(ExecuteNonQuery.AffectedRecords)];
+            Assert.Equal(0, affected);
```
alexandru-petre, line 337: With the guard in place nothing ever writes this entry, so the 0 comes from the out argument's default initialisation inside `WorkflowInvoker` rather than from anything the activity did. The assertion passes identically whether the activity writes 0 or writes nothing at all.

The test is still a genuine regression test: without the fix `Invoke` throws `NullReferenceException` and the test fails before reaching this line, so the verification that matters does happen. The issue is that this line reads as though it pins down an output contract when it does not, and the "writes nothing" case is the one that leaves a stale value in a real workflow (see the comment on `ExecuteNonQuery.cs`).

Suggestion: bind the output to a pre-seeded variable so the semantics is actually pinned down. Wrap the activity in a `Sequence` carrying a `Variable<int> { Default = 42 }`, bind `AffectedRecords` to it, and read it back afterwards; `CaptureConnection` further down this file is a ready made template for the read-back activity. Then assert 42 if stale values are intended, or 0 if the reset is.

alexandru-petre (commented):
Reviewed the branch against STUD-81504. The fix is the guard the ticket prescribed, in the file it named, and the new test drives the reported repro, so the stated intent and the implementation agree and I found nothing wrong with the change itself.

Two comments below, the first one would require a decision before the merge. Both are about output semantics that neither the ticket nor the `ContinueOnError` property documentation ever states, rather than about defects in the fix.

Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
viogroza: Fixed in 285b72c: `affectedRecords` is now initialized to `new DBExecuteCommandResult()` instead of `null`, so on the swallowed-failure path `AffectedRecords` resets to `0` — matching `InsertDataTable`/`BulkInsert`/`BulkUpdate`. The now-unreachable null guard is removed.

On the `Out`/`InOut` `Parameters` point: those remain untouched on this path (`ParametersBind` is empty, so `SetOutputParameters` has nothing to write), which is the same behavior `ExecuteQuery` already has for its own `Parameters` output on a swallowed failure. Giving arbitrary-typed Out/InOut parameters a well-defined "reset" value felt like a bigger design decision than this fix's scope (there's no natural default the way `0` is for `AffectedRecords`), so I left that as-is rather than guessing. Happy to open a follow-up ticket if you'd like that pinned down explicitly too.

Activities/Database/UiPath.Database.Tests/SqliteIntegrationTests.cs
viogroza: Added a new test in 285b72c — `ExecuteNonQuery_ContinueOnError_ResetsAffectedRecordsInsteadOfLeavingStaleValue` — using the pre-seeded-variable pattern you suggested: wraps the activity in a `Sequence` with a `Variable<int> { Default = 42 }` bound to `AffectedRecords`, reads it back with a generic `CaptureValue<T>` capture activity (generalized from `CaptureConnection`), and asserts it resets to `0`. Since `AffectedRecords` now genuinely resets to `0` (see the reply on the `ExecuteNonQuery.cs` comment), this assertion is real, not incidental to `WorkflowInvoker`'s default. Kept the original test as well since it still validates the crash regression itself.

Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
```diff
@@ -47,7 +47,7 @@ protected async override Task<Action<AsyncCodeActivityContext>> ExecuteAsync(Asy
                 string sql = string.Empty;
                 int? commandTimeoutMs = TimeoutMS.Expression is null ? (int?)null : TimeoutMS.Get(context);
                 DatabaseConnection existingConnection = null;
-                DBExecuteCommandResult affectedRecords = null;
+                DBExecuteCommandResult affectedRecords = new DBExecuteCommandResult();
```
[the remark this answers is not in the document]
viogroza: Correct catch — the PR description was stale, not the code. This was a deliberate change, not an accidental one: per review discussion on this PR (see the two `alexandru-petre` threads above), leaving `AffectedRecords` untouched/stale on a swallowed failure was judged worse than resetting it to `0`, since a stale value silently reports a prior run's row count as if the failed statement succeeded (e.g. in a loop). `InsertDataTable`/`BulkInsert`/`BulkUpdate` already reset to `0` on a swallowed error, so this makes `ExecuteNonQuery` consistent with those rather than with `ExecuteQuery`'s current stale-on-failure behavior, which isn't itself validated as correct.

`Out`/`InOut` `Parameters` are intentionally NOT included in this reset — `ParametersBind` stays empty on the failure path regardless, so those remain untouched, since there's no natural default for an arbitrary parameter type the way `0` is for an int.

Updated the PR description to describe this explicitly instead of the outdated "matches ExecuteQuery" framing.

alexandru-petre (approved):
The only open item is Copilot's comment on the `TimeoutMS` exception

Activities/Database/UiPath.Database.Activities/ExecuteNonQuery.cs
```diff
@@ -47,10 +47,10 @@ protected async override Task<Action<AsyncCodeActivityContext>> ExecuteAsync(Asy
...
                 if (commandTimeoutMs.HasValue && commandTimeoutMs.Value < 0)
                 {
-                    throw new ArgumentException(Resources.TimeoutMSException, nameof(TimeoutMS));
+                    throw new ArgumentException(Resources.TimeoutMSException);
```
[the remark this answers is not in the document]
viogroza: Fair catch — reverted in c7c6628: restored `nameof(TimeoutMS)` as the `ArgumentException` paramName. Dropping it only here (to satisfy CA2208) made `ExecuteNonQuery` inconsistent with `ExecuteQuery.cs:59`, `BulkInsert.cs:62`, `InsertDataTable.cs:60`, and `BulkUpdate.cs:76`, which all still pass it and weren't touched by this PR. Fixing the underlying Sonar finding pack-wide is out of scope here; left as-is to match the siblings. The new negative-`TimeoutMS` regression test is unaffected (it only asserts the exception type).

